PR #11, opened 2019-09-18
Modify Image Creation

chases2:
Modify images so that the built executable sits in the PWD

Also remove config-upload.sh; it's not part of any automated build process, it's a mediocre example of `Configurator` usage, and it doesn't work in this repo anymore


cmd/configurator/BUILD.bazel
```diff
@@ -9,7 +9,12 @@ go_binary(
     pure = "on",
 )
 
-make_image(name = "image")
+make_image(
+    name = "image",
+    directory = "/",
+    files = [":configurator"],
```
fejta, line 15: Why? This isn't what we should be doing and this isn't how make_image works (note that it embeds :go_default_library and builds it differently.

This breaks macos builds.

Issue #13, opened 2019-09-19
Configurator swallows some GCS errors


chases2:
See https://github.com/kubernetes/test-infra/pull/14367

PR #16, opened 2019-09-19
Remove "Configurator" from TestGrid Repository

chases2:
As a Prow-dependent utility, it should stay with Prow.
In the future, an agnostic version of Configurator can be created as need arises

This leaves `config.proto` as the interface that users of TestGrid should adhere to for now. Since all of TestGrid's current users have some method of generating this proto, this shouldn't affect usability in the short term. More general tools and libraries can be built as open-sourcing continues.


chases2:
Also closes #13 (a vacuous truth)

PR #18, opened 2019-09-25
Move protos into one folder, don't rely on .pb.go

michelle192837:
There's some issues with Bazel when proto imports aren't explicit (like with config.go and config.pb.go being in the same Go package). Moving protos into another folder to fix this.


michelle192837:
PTAL?

michelle192837:
Testing
/meow

PR #20, opened 2019-09-27
Removing conformance files, they're Kubernetes-specific.

michelle192837:


conformance/upload_e2e.py
```diff
@@ -1,328 +0,0 @@
...
-        '--key-file',
-        help='path to GCP service account key file, which will be activated before '
-        'uploading if provided, the account will be revoked and the active account reset '
-        'on exit',
-        required=False,
-    )
-    return parser.parse_args(args=cli_args)
-
-
-def main(cli_args):
```
fejta: This is potentially a useful way for people to figure out how to upload job results when not using prow.

BenTheElder: thanks for the heads up -- as long as this continues to live on in test-infra for existing users everything should be fine.

we might want to bring it or something similar back here for future users though 😅

Issue #26, opened 2019-10-08
Use consistent protos everywhere


michelle192837:
Some of these are already handled, some are partially exposed, some we don't have a ready solution for (config.proto). These are the formats that TestGrid uses throughout the updaters/frontend, so we want to update them in one place instead of porting new changes out inconsistently.

As of November 12, 2019:
- [ ] config.proto
  - [x] ...is in GCP/testgrid
  - [ ] ...is the source of truth
- [x] issue_state.proto
  - [x] ...is in GCP/testgrid
  - [x] ...is the source of truth
- [ ] state.proto
  - [x] ...is in GCP/testgrid
  - [ ] ...is the source of truth
- [x] summary.proto
  - [x] ...is in GCP/testgrid
  - [x] ...is the source of truth
- [x] test_status.proto
  - [x] ...is in GCP/testgrid
  - [x] ...is the source of truth
- [x] custom_evaluator.proto
  - [x] ...is in GCP/testgrid
  - [x] ...is the source of truth

fejta:
This one still WIP?

michelle192837:
Yup, config.proto is being a jerk. We could probably switch state.proto over at least.

PR #27, opened 2019-10-08
Updating the issue_state and test_status protos.

michelle192837:
Updating/adding these to align with the actual protos so we can use the same one everywhere.

xref #26


michelle192837:
Done I think?

/hold cancel

Issue #7, opened 2019-09-16
Configurator accepts only one default file


chases2:
Configurator parses through a directory of files, looking for YAML configuration files to populate a config.proto. Configurator currently accepts an optional `--default` file that it uses as a default for the entire tree. Defaults are otherwise ignored.

It may be more useful for Configurator to recognize default settings (since they use different YAML keys already) and apply that default to this directory and all subdirectories, but _not_ to other directories in the tree.

This enhancement may require replacing the os.Walk function with a recursive solution.

chases2:
I think this is a solved issue for configurator. In addition, the move away from transfigure has made this utility less useful in-general. Closing this issue.

PR #34, opened 2019-10-14
Refactoring of Config Libraries to include Yaml Config

chases2:
Includes defaulting behaviour while deprecating that struct from config.proto, for later deletion.

Incremental progress towards #26 and #7


chases2:
Update: includes `MarshalYAML` function in YAML library.
Unmarshalling is covered with `Update` in the same library.

PR #35, opened 2019-10-16
Additions to config.proto and custom_evaluator.proto

chases2:
Additional work on #26


pb/config/config.proto
```diff
@@ -46,20 +55,24 @@ message Notification {
...
+  // Path to the test result location, including protocol.
+  // The location must be accessible from the Update Server, and not local.
+  // GCS is usually used, making this field "gs://<bucket>"
   string gcs_prefix = 2;
 
   // Number of days of test results to gather and serve.
   int32 days_of_results = 3;
 
-  // Whether to ignore pending (currently running) test results.
+  // Whether to ignore pending (currently running) test results. Recommended.
```
michelle192837: "Recommended" as what? (Personally I'd recommend 'ignore_pending=false' usually since you can link to pending results).

pb/config/config.proto
```diff
@@ -72,6 +85,14 @@ message TestGroup {
   // ignored, then the name of the tests will be the build target.
   TestsName tests_name_policy = 6;
 
+  // Whether to infer the test properties from the test results.
+  // Experimental; recommend "false"
+  bool gather_test_properties = 7;
```
michelle192837: This should be 'true' generally (and also we should delete it eventually).

pb/config/config.proto
```diff
@@ -72,6 +85,14 @@ message TestGroup {
...
+
+  // Tests with names that include these substrings will be removed from the
+  // table.
```
michelle192837: 'from state' actually (before we create a table)

pb/config/config.proto
```diff
@@ -89,13 +125,48 @@ message TestGroup {
   // a consistent failure. If zero, no alerts are raised.
   int32 num_failures_to_alert = 12;
 
+  // Whether to automatically file bugs. Requires further implementation of additional components.
+  int32 bug_component = 13;
```
michelle192837: And the component they will be filed to.

pb/config/config.proto
```diff
@@ -89,13 +125,48 @@ message TestGroup {
...
+
   // Default code search path for searching regressions. Overridden by
   // code_search_path in DashboardTab.
   string code_search_path = 14;
 
   // The number of columns to consider "recent" for a variety of purposes.
   int32 num_columns_recent = 15;
 
+  // Whether to read test metadata from the test results. Information
+  // from the test metadata is used to open bugs in specific cases.
```
michelle192837: To route bugs when filed.

Issue #38, opened 2019-10-21
Configure bazel jobs to use RBE


fejta:
This does't look right:

https://github.com/GoogleCloudPlatform/testgrid/blob/9f0319fc11974c08ebb9b58044d25fdb70d1ba44/.bazelrc#L69

PR #40, opened 2019-10-23
Fixup .bazelrc

chases2:
Tested `bazel build //...` and `bazel test //...` both remotely and locally, using Bazel v1.1.0

Removing old build-ci settings since GCP/oss-test-infra Prow doesn't use RBE, and even if it did, the configuration should sit with the Prow instance and not here.

Removing `--host_force_python` since there's no python code in this repository

Also removing `--incompatible_disable_deprecated_attr_params` since we don't seem to be using deprecated params either.

/assign @fejta 

Addresses #38

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -2,15 +2,6 @@
 build --workspace_status_command=./hack/print-workspace-status.sh
 run --workspace_status_command=./hack/print-workspace-status.sh
 
-# https://github.com/kubernetes/test-infra/issues/13140
-build --incompatible_disable_deprecated_attr_params=false
-query --incompatible_disable_deprecated_attr_params=false
-
-# https://github.com/kubernetes/test-infra/issues/13239
-# https://github.com/bazelbuild/rules_docker/issues/842
-# https://github.com/bazelbuild/bazel/issues/7899
-build --host_force_python=PY2
-
 # enable data race detection
 test --features=race --test_output=errors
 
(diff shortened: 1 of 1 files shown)


.bazelrc
```diff
@@ -2,15 +2,6 @@
 build --workspace_status_command=./hack/print-workspace-status.sh
```
fejta: Why do we need this?

.bazelrc
```diff
@@ -2,15 +2,6 @@
...
-build --incompatible_disable_deprecated_attr_params=false
-query --incompatible_disable_deprecated_attr_params=false
-
-# https://github.com/kubernetes/test-infra/issues/13239
-# https://github.com/bazelbuild/rules_docker/issues/842
-# https://github.com/bazelbuild/bazel/issues/7899
-build --host_force_python=PY2
-
 # enable data race detection
 test --features=race --test_output=errors
```
fejta: Why do we need the --build_tests_only?

.bazelrc
```diff
@@ -25,8 +16,8 @@ test:lint --test_tag_filters=lint
 # bazel test //... --config=unit
 test:unit --test_tag_filters=-lint
```
fejta: Can we drop this too?

.bazelrc
```diff
@@ -45,31 +36,21 @@ build:remote --define=EXECUTOR=remote
...
+build:inmemory --experimental_inmemory_jdeps_files
+build:inmemory --experimental_inmemory_dotd_files
 
-# --config=ci-instance adds the instance name
-build:ci-instance --remote_instance_name=projects/k8s-prow-builds/instances/default_instance
+build:toplevel --config=inmemory
+build:toplevel --experimental_remote_download_outputs=toplevel
 
-# Config we want to use in ci
-build:ci --config=remote --config=ci-instance
```
fejta: Can we please keep a `build:ci` option, even if it does nothing? That way we can (continue) to define `bazel test --config=ci` jobs, and just change what this means over time.

`build:ci --noshow_progress` is a good value to add for non-interactive jobs

PR #41, opened 2019-10-23
Update state proto with additional fields.

michelle192837:


michelle192837:
Going to clean up the comments for consistency, I'll remove WIP when that's done.

I expect we'll need to rename some fields again but will tackle that separately.

michelle192837:
Still can't get Bazel to find the timestamp proto :/ So leaving it there and reserving the fields for now.

pb/state/state.proto
```diff
@@ -40,21 +136,128 @@ message Row {
...
+  reserved 6;
+
+  // Names of metrics associated with this test case. Stored separate from
+  // metric info (which may be omitted).
+  repeated string metric = 7;
 
   repeated Metric metrics = 8;  // Numerical performance/timing data, etc.
 
-  repeated string icons = 9;  // Put 1-2 chars inside cell (F for fail, etc)
+  repeated string icons = 9;  // 1-5 chars inside cell (F for fail, etc)
```
fejta: woo!

pb/state/state.proto
```diff
@@ -40,21 +136,128 @@ message Row {
...
+
+  // Summaries of dashboard tabs associated with this group.
+  repeated DashboardTabSummary dashboard_tab_summaries = 3;
+
+  // Metrics associated with the update for this group.
+  UpdateInfo update_entry = 4;
+}
+
+// An updater server updates test groups upon request.
+service Updater {
```
fejta: I suspect this is going to import a lot of unnecessary junk. Thoughts on moving update RPCs into their own proto?

michelle192837: That's reasonable yeah, let me go do that.

michelle192837: Done!

fejta:
Maybe create an alpha.2 release if you start pulling this code into somewhere else (k/ti, etc) https://github.com/GoogleCloudPlatform/testgrid/releases

michelle192837:
Noted! Not yet yeah, but probably not far off.

PR #42, opened 2019-10-23
Also fixing .bazelrc

chases2:
Addressing previous comments from #40 

The print-workspace-status info is still here because that hack script is present and maintained by this repository to push images of Testgrid components. See [`//images:push'](https://github.com/GoogleCloudPlatform/testgrid/blob/master/images/BUILD.bazel)

The rest of the suggestions have been taken.

/assign @fejta


chases2:
xref issue #38

PR #48, opened 2019-11-13
Create state based summarizer

fejta:
* Download the config proto
* For each dashboard create a summary for each tab
* For each tab
  - Download the state proto
  - Filter out irrelevant rows
  - Find the latest run, noting if this is stale (also last update to state proto)
  - Extract any active alerts
  - Find the last result with only green tests
  - Decide the overall tab state:
     * stale if results are old
     * failing if there are active alerts
     * passing if all recent results are healthy
     * flaky if there are some failures but no alerts
     * unknown if there are not any results yet

TODO:
* Many more unit tests (let me know if you want logic to come with tests or added later)
* Failing/passing alert timestamps (external proto doesn't yet define these fields)
* Support for functionality that isn't yet being used by testgrid.k8s.io
* Performance improvements:
  - More concurrency, similar to the updater
  - Only update when the state has changed
* Upload the summary protos after creating them

### DIFF
--- a/cmd/updater/main.go
+++ b/cmd/updater/main.go
@@ -375,6 +375,7 @@ func AppendColumn(headers []string, format nameConfig, grid *state.Grid, rows ma
 				r = &state.Row{
 					Name: name,
 					Id:   target,
+					// TODO(fejta): AlertInfo,
 				}
 				rows[name] = r
 				grid.Rows = append(grid.Rows, r)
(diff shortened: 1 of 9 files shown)


cmd/updater/main.go
```diff
@@ -375,6 +375,7 @@ func AppendColumn(headers []string, format nameConfig, grid *state.Grid, rows ma
 				r = &state.Row{
 					Name: name,
 					Id:   target,
+					// TODO(fejta): AlertInfo,
```
fejta: I think I see some issues with how we open/close these alerts, but I eventually realized the updater is what is responsible for this, not the summarizer. So any issues are not really relevant here, added a TODO to move this logic into go

PR #54, opened 2019-11-18
Add fail_time and pass_time to AlertInfo.

fejta:
A super hacky way to inject the source protos for well-known types. Need to figure out something better soon.


michelle192837:
I'm impressed you got this working, even if it's hacky XD

PR #55, opened 2019-11-19
Make updater configure alerts.

fejta:
Also make summarizer set pass/fail time.
Encapsulate result logic into an internal package (so both update and summarizer can use it).

### DIFF
--- a/cmd/updater/main.go
+++ b/cmd/updater/main.go
@@ -24,6 +24,7 @@ import (
 	"flag"
 	"fmt"
 	"log"
+	"math"
 	"net/url"
 	"path"
 	"runtime"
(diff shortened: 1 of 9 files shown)


fejta:
Passing tests now (woops)!

cmd/updater/main.go
```diff
@@ -402,6 +404,111 @@ func AppendColumn(headers []string, format nameConfig, grid *state.Grid, rows ma
...
+	defer cancel()
+	var failures int
+	var totalFailures int32
+	var passes int
+	var compressedIdx int
+	ch := result.Iter(ctx, row.Results)
+	var lastFail *state.Column
+	var latestPass *state.Column
+	var failIdx int
+	for _, col := range cols {
```
michelle192837: nit: Function looks good! Some extra comments would be great.

fejta: Done

cmd/updater/main_test.go
```diff
@@ -325,3 +326,238 @@ func TestMarshalGrid(t *testing.T) {
...
+				Results: []int32{
+					int32(state.Row_PASS), 1,
+					int32(state.Row_FAIL), 5,
+				},
+			},
+			failOpen: 1,
+		},
+		{
+			name: "track through skipped results",
+			row: state.Row{
```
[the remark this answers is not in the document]
fejta: Done

fejta:
consecutive has a C
consecutive has a C
consecutive has a C...

PR #57, opened 2019-11-22
Create summarizer deployment.

fejta:
Also:
* Add a -wait flag to summarizer and updater, which makes them run forever.
* Rename make_image to go_image
* Build all targets in CI during bazel test //...

### DIFF
--- a/cmd/updater/main.go
+++ b/cmd/updater/main.go
@@ -60,6 +60,7 @@ type options struct {
 	group            string
 	groupConcurrency int
 	buildConcurrency int
+	wait             time.Duration
 }
 
 // validate ensures sane options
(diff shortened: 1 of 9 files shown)


cmd/updater/main.go
```diff
@@ -89,6 +90,7 @@ func gatherOptions() options {
 	flag.StringVar(&o.group, "test-group", "", "Only update named group if set")
 	flag.IntVar(&o.groupConcurrency, "group-concurrency", 0, "Manually define the number of groups to concurrently update if non-zero")
 	flag.IntVar(&o.buildConcurrency, "build-concurrency", 0, "Manually define the number of builds to concurrently read if non-zero")
+	flag.DurationVar(&o.wait, "wait", 0, "Ensure at least this much time has passed since the last loop (exit if zero).")
```
michelle192837: Example flag argument (like '5m'?)

fejta: Meh, we don't do that elsewhere, they can see how we're using it, `--help` says it is a duration, and its a well-known go pattern.

Can fix later if necessary.

/hold cancel

michelle192837: Thought it might be, and good point. 👍

PR #59, opened 2019-11-22
Allow reading from beta subdir

fejta:
* Use service account in summarizer, updater
* Update immediately the first time

### DIFF
--- a/cmd/summarizer/main.go
+++ b/cmd/summarizer/main.go
@@ -56,8 +56,8 @@ func (o *options) validate() error {
 	if o.config.String() == "" {
 		return errors.New("empty --config")
 	}
-	if o.config.Bucket() == "k8s-testgrid" && o.confirm { // TODO(fejta): remove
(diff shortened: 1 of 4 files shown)


cmd/summarizer/main.go
```diff
@@ -56,8 +56,8 @@ func (o *options) validate() error {
 	if o.config.String() == "" {
 		return errors.New("empty --config")
 	}
-	if o.config.Bucket() == "k8s-testgrid" && o.confirm { // TODO(fejta): remove
-		return fmt.Errorf("--config=%q cannot start with gs://k8s-testgrid", o.config)
+	if o.config.Bucket() == "k8s-testgrid" && o.config.Object() == "config" && o.confirm { // TODO(fejta): remove
```
michelle192837: Can this just prevent anything that isn't under the 'beta' path? (Or alternatively, test with k8s-testgrid-canary?)

fejta: Done

PR #61, opened 2019-11-25
Improve updater, summarizer, gcs reader

fejta:
* Use its own bucket rather than a subdir
* Support running continuously if `--wait` is positive
* Update all dashboards even if one fails
* Switch updater to use logrus
* Switch updater to use the configured number of days of results rather than 7
* Advance compressed index on running cells.
* Sort builds from the presubmit directory correctly.

Do you want to break this up?

### DIFF
--- a/cmd/updater/main.go
+++ b/cmd/updater/main.go
@@ -23,7 +23,6 @@ import (
 	"errors"
 	"flag"
 	"fmt"
-	"log"
 	"math"
 	"net/url"
 	"path"
@@ -43,6 +42,7 @@ import (
 	"cloud.google.com/go/storage"
 	"github.com/golang/protobuf/proto"
 	"github.com/golang/protobuf/ptypes/timestamp"
+	"github.com/sirupsen/logrus"
 	"vbom.ml/util/sortorder"
 )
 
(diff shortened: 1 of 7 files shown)


cmd/updater/main.go
```diff
@@ -828,7 +845,11 @@ func ReadBuilds(parent context.Context, group configpb.TestGroup, builds Builds,
 		appendColumn(grid, heads, nameCfg, rows, *c)
 		alertRows(grid.Columns, grid.Rows, failsOpen, passesClose)
 		if c.Started < stop.Unix() { // There may be concurrency results < stop.Unix()
-			log.Printf("  %s#%s before %s, stopping...", group.Name, c.ID, stop)
+			logrus.WithFields(logrus.Fields{
+				"group": group.Name,
+				"id":    c.ID,
+				"stop":  stop,
+			}).Info("Group started before latest time, stopping")
```
michelle192837: What case is this handling, out of curiosity?

cmd/updater/main.go
```diff
@@ -515,8 +526,8 @@ func stamp(col *state.Column) *timestamp.Timestamp {
 
 const elapsedKey = "seconds-elapsed"
 
-// ReadBuild asynchronously downloads the files in build from gcs and convert them into a build.
-func ReadBuild(build Build) (*Column, error) {
+// readBuild asynchronously downloads the files in build from gcs and convert them into a build.
```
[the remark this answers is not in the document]
fejta: Done

cmd/updater/main.go
fejta: Done

stops processing earlier columns.

PR #62, opened 2019-11-27
Simplify gcs and updater concurrency

fejta:
* Remove context from gcs.Build, instead require caller to pass one in.
* Ensure updater closes all channels
* Timeout group update, make both and build timeout configurable by flag.
* Fix a potential race where multiple workers encounter an error and try to write to the error channel
* Simplify how we get started and finished
* Simplify how we cancel reading builds
* Add a --debug flag, drop intra-group logging to debug
* Add a periodic progress log about update
* Use x-goog-meta-link instead of link (server no longer returns the latter)

### DIFF
--- a/cmd/updater/main.go
+++ b/cmd/updater/main.go
@@ -57,10 +57,13 @@ type options struct {
 	config           gcs.Path // gs://path/to/config/proto
 	creds            string
 	confirm          bool
+	debug            bool
 	group            string
 	groupConcurrency int
 	buildConcurrency int
 	wait             time.Duration
+	groupTimeout     time.Duration
+	buildTimeout     time.Duration
 }
 
 // validate ensures sane options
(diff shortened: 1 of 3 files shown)


cmd/updater/main.go
```diff
@@ -934,15 +951,46 @@ func updateOnce(ctx context.Context, opt options) {
 		}
```
michelle192837: nit: Are the commented lines (943-946) supposed to be there?

cmd/updater/main.go
```diff
@@ -934,15 +951,46 @@ func updateOnce(ctx context.Context, opt options) {
...
 		groups <- *tg
 	} else { // All groups
-		for _, tg := range cfg.TestGroups {
+		start := time.Now()
```
michelle192837: There's a few long functions in this file (though the comments help a lot in tracking what they're doing), consider breaking them up?

PR #69, opened 2019-12-02
Add/Rename Config.proto fields

chases2:
Renames TESTS_NAME_MIN to TESTS_NAME_UNSPECIFIED (see [style guide](https://developers.google.com/protocol-buffers/docs/style#enums))

Adds additional fields

/assign @fejta @michelle192837


chases2:
Thanks for finding the reference to the changed enum name, robot friend

PR #70, opened 2019-12-02
Invert dashboard option for failing tabs

chases2:
Proto3 defaults to false, and this default behavior is preferable

/assign @fejta @michelle192837


pb/config/config.proto
```diff
@@ -427,7 +427,8 @@ message Dashboard {
   // Defaults to Summary
   string default_tab = 5;
 
-  bool highlight_failing_tabs = 6;
+  // Controls whether to suppress highlighting of failing tabs.
+  bool downplay_failing_tabs = 6;
```
michelle192837: Ah, not sure on this name, mostly because it sounds like a partial disable and not a full/clear one? (Don't think 'understate' works either). But not sure on a better one, so -shrug- replace it if you think of better!

fejta: Why do we need/want to reuse field 6 here?

michelle192837: Shoot, missed that, thanks! Yeah, we probably want to make and then migrate code and config to a new field if we're going to do this.

chases2: It seemed cleaner to me to have field 6 deal with the same option, even if its default is reversed.

However, I don't think that's the Proto Way of migrating; fields are easier and safer to add, deprecate, and reserve, so let's do that.

chases2:
Updated to deprecate instead of switch
/hold cancel

fejta:
It's not just the default right? Anyone who sets field 6 to true will soon get the opposite behavior

PR #72, opened 2019-12-12
Write summaries to gs://k8s-testgrid/summary-foo

fejta:


fejta:
//cc @cjwagner

cjwagner:
Hmmm, I am an org member.

PR #76, opened 2020-01-04
Add the properties field for the Target struct

tony-yang:
Add the properties field for the Target struct to match the resultstore's Target proto.


tony-yang:
> Can/should this have a unit test?

Thanks Erick. Unit test is added in PR 80

PR #86, opened 2020-01-15
Remove apimachinery dependency

fejta:
We only use it for testing, we can remove a lot of dependencies by refactoring that file


fejta:
Passing tests now!

PR #87, opened 2020-01-16
Update config.proto with different ColumnHeader options

chases2:
Working towards #26


pb/config/config.proto
```diff
@@ -61,7 +61,7 @@ message TestGroup {
   // Path to the test result location, including protocol.
   // The location must be accessible from the Update Server, and not local.
   // GCS is usually used, making this field "gs://<bucket>"
-  string gcs_prefix = 2;
```
fejta: So this breaks all open source usage of testgrid.

michelle192837: Is this actively breaking things, or breaking upgrading other repos'/test-infra's upgrade?

michelle192837: Ah nvm, answered in #103

fejta: Someone doing a patch upgrade of their testgrid repo will hit this. IMO we should just patch it internally when/if we want to use a different name for this field.

michelle192837: I'm not sure that's worth it long-term since gcs_prefix is a very specific name for the field anyways (though I'm fine with it short-term to get test-infra updated). Can I just change the usage of this in test-infra afterwards?

michelle192837: (along with making a new minor-version change I assume, to avoid hitting this when patching?)

PR #95, opened 2020-02-12
Validate TestGrid config uses unique names.

michelle192837:
Add some universal validation of the config proto to TestGrid.

Will be following up with more validation in a follow-up PR.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -22,6 +22,7 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"regexp"
 	"strings"
 
 	"cloud.google.com/go/storage"
@@ -29,6 +30,7 @@ import (
 
 	configpb "github.com/GoogleCloudPlatform/testgrid/pb/config"
 	"github.com/GoogleCloudPlatform/testgrid/util/gcs"
+	multierror "github.com/hashicorp/go-multierror"
 )
 
 // MissingFieldError is an error that includes the missing field.
@@ -40,8 +42,43 @@ func (e MissingFieldError) Error() string {
 	return fmt.Sprintf("field missing or unset: %s", e.Field)
 }
 
-// Validate checks that a configuration is well-formed, having test groups and dashboards set.
+// DuplicateNameError is an error that includes the duplicate name.
+type DuplicateNameError struct {
+	Name string
+}
+
+func (e DuplicateNameError) Error() string {
+	return fmt.Sprintf("found duplicate name after normalizing: %s", e.Name)
+}
+
+// normalize lowercases, and removes all non-alphanumeric characters from a string.
+func normalize(s string) string {
+	regex := regexp.MustCompile("[^a-zA-Z0-9]+")
+	s = regex.ReplaceAllString(s, "")
+	s = strings.ToLower(s)
+	return s
+}
+
+// validateUnique checks that a list has no duplicate entries.
+func validateUnique(items []string) error {
+	mErr := &multierror.Error{}
+	set := map[string]bool{}
+	for _, s := range items {
+		_, ok := set[s]
+		if ok {
+			mErr = multierror.Append(mErr, DuplicateNameError{s})
+		} else {
+			set[s] = true
+		}
+	}
+	return mErr.ErrorOrNil()
+}
+
+// Validate checks that a configuration is well-formed.
 func Validate(c configpb.Configuration) error {
+	mErr := &multierror.Error{}
+
+	// TestGrid requires at least 1 TestGroup and 1 Dashboard in order to do anything.
 	if len(c.TestGroups) == 0 {
 		return MissingFieldError{"TestGroups"}
 	}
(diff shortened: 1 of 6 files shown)


config/config.go
```diff
@@ -40,16 +43,92 @@ func (e MissingFieldError) Error() string {
...
+// Normalize lowercases, and removes all non-alphanumeric characters from a string.
+func Normalize(s string) string {
+	regex := regexp.MustCompile("[^a-zA-Z0-9]+")
+	s = regex.ReplaceAllString(s, "")
+	s = strings.ToLower(s)
+	return s
+}
+
+// ValidateUnique checks that a list has no duplicate entries.
+func ValidateUnique(items []string) error {
```
fejta: Why do you need to export this function and all the others?

config/config.go
```diff
@@ -40,16 +43,92 @@ func (e MissingFieldError) Error() string {
...
+	mErr := &multierror.Error{}
+	set := sets.String{}
```
fejta: Why don't you just use a dictionary and not bloat the code?
```golang
var errors []error
foo := map[string]bool{}
if _, present := foo[s]; present {
  errors = append(errors, DuplicateNameError{s})
}
foo[s] = true
return fmt.Errorf("%d errors: %s")
```

config/config.go
michelle192837: Don't need to, changed!

config/config.go
michelle192837: Replaced use of sets, I think multierror should be okay and that the convenience methods are worth it.

config/config.go
```diff
@@ -40,16 +42,94 @@ func (e MissingFieldError) Error() string {
...
+			set[s] = true
+		}
+	}
+	return mErr.ErrorOrNil()
+}
+
 // Validate checks that a configuration is well-formed, having test groups and dashboards set.
+// Also checks each group of entities (TestGroup, DashboardTab, Dashboard, DashboardGroup) for
+// unique names and valid options.
 func Validate(c configpb.Configuration) error {
```
chases2: Adding comments throughout this function describing why we're failing validation on this condition would help with its readability.

In that way, instead of updating the comment above for all the things that Validate updates, you can look through /add to the code. We're likely to refactor and extend this method.

config/config.go
```diff
@@ -40,16 +42,94 @@ func (e MissingFieldError) Error() string {
...
+	mErr := &multierror.Error{}
+
 	if len(c.TestGroups) == 0 {
```
chases2: Ex. `TestGrid needs Test Groups and Dashboards to display anything.`

config/config.go
```diff
@@ -40,16 +42,94 @@ func (e MissingFieldError) Error() string {
...
 	if len(c.Dashboards) == 0 {
 		return MissingFieldError{"Dashboards"}
 	}
 
-	return nil
+	tgNames := []string{}
+	for _, tg := range c.TestGroups {
+		tgNames = append(tgNames, normalize(tg.GetName()))
+	}
+	err := validateUnique(tgNames)
```
chases2: Ex. `Test Group names must be unique after being normalized.`

config/config.go
```diff
@@ -40,16 +42,94 @@ func (e MissingFieldError) Error() string {
...
+		for _, tab := range dash.GetDashboardTab() {
+			tabNames = append(tabNames, normalize(tab.GetName()))
+		}
+		// Dashboard tab names must be unique within a dashboard.
+		err = validateUnique(tabNames)
+		if err != nil {
+			mErr = multierror.Append(mErr, err)
+		}
+	}
+	err = validateUnique(dashNames)
```
chases2: Ex. `Two dashboards with the same name should be combined into one`

config/config_test.go
```diff
@@ -69,12 +100,88 @@ func TestUpdate_Validate(t *testing.T) {
...
+			},
+			expectedDup: map[string]bool{"name1": true},
+		},
 	}
 
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
 			err := Validate(test.input)
-			if err != nil && test.expectedMissing == "" {
+			if err != nil && test.expectedMissing == "" && len(test.expectedDup) <= 0 {
```
chases2: Prefer `len(test.expectedDup) == 0`. I'm pretty sure there's no situation in which len return a negative number, and I've already flipped the < twice in my head.

config/config_test.go
```diff
@@ -85,6 +192,22 @@ func TestUpdate_Validate(t *testing.T) {
 					t.Errorf("Expected MissingFieldError(%s), but got %v", test.expectedMissing, err)
 				}
 			}
+
+			if len(test.expectedDup) > 0 {
+				if err == nil {
+					t.Errorf("Expected DuplicateNameError(%s), but got no error", test.expectedDup)
```
chases2: I'm surprised you can serialize a `map[string]bool` with %s instead of %v

config/config.go
michelle192837: Yup! Clarified some, does this work?

config/config.go
michelle192837: Done

config/config.go
michelle192837: Done

michelle192837:
/hold cancel

Thanks!

PR #97, opened 2020-02-12
Refactor cmd/updater to import a pkg/updater library

chases2:
xref #91

Also:
- Fixes a nit in summary_test.go
- Emits data about overall update times to `Info`


chases2:
> Thoughts on splitting this into 2 commits (move to new dir, refactor both files)

Most of the refactoring (excepting the two bullet points mentioned above) is because of the file moving to a new directory. Splitting them doesn't seem worth the effort, and if there's an issue I'm fine with reverting the whole thing.

chases2:
Update: split into separate commits

Issue #98, opened 2020-02-14
Add full config validation


michelle192837:
We have barebones validation of config.proto at the moment, but there are a lot of universal checks that config.protos anywhere have to pass for TestGrid to use them (or to avoid unintuitive behavior). 

- [x] No duplicate entities allowed (e.g. no Test Groups with the same post-normalization name)
- [x] Referenced entities must exist (e.g. the Test Group a Dashboard Tab is backed by must exist)
- [x] Configured entities in TestGrid should show up somewhere (e.g. no defining a Test Group without at least 1 corresponding Dashboard Tab)
- [x] Test Group options should be reasonable
- [x] Dashboard Tab options should be reasonable
- [ ] Dashboard options should be reasonable
- [x] Dashboard Group options should be reasonable (will be covered by the first three checks, mentioned for completeness)

chases2:
Validate currently checks a wide variety of TestGroup options in this function:https://github.com/GoogleCloudPlatform/testgrid/blob/1844fdad5f0b8ddc2101e52069cca0209403e765/config/config.go#L246
Same for DashboardTabs:(https://github.com/GoogleCloudPlatform/testgrid/blob/1844fdad5f0b8ddc2101e52069cca0209403e765/config/config.go#L364)

It might be worth looking to see if Dashboards need similar checks to be added, but I'm going to check off the "should be reasonable" requirements for the first two.

michelle192837:
Thanks Sean! I agree, I think we have pretty reasonable validation all around (Dashboards really don't have too much to check)

chases2:
Looking at the [dashboard proto](https://github.com/GoogleCloudPlatform/testgrid/blob/1844fdad5f0b8ddc2101e52069cca0209403e765/pb/config/config.proto#L444), I think it might be worthwhile to enforce the following:
[ ] Dashboards must contain at least one DashboardTab
[ ] Dashboard names must be at least 3 characters long after normalization (Naming your dashboard, for example, "q" will result in it not being displayable in the UI)

Then I think this issue can be closed. Thoughts?

michelle192837:
I think that's fair, yeah. First seems like a reasonable way to avoid incorrect configs, second is a good way to avoid collisions with the web endpoints. 👍 !

PR #99, opened 2020-02-19
Validate referenced entities exist in config.

michelle192837:
Also adds more useful, structured errors.

ref #98


fejta:
/uncc @chases2 
Deferring to Sean, add me back if you want me to look at it.

PR #101, opened 2020-02-27
Support summarizing non kubernetes clients.

fejta:
Also stop assuming an extra column exists, even when processing a kubernetes client.


pkg/summarizer/summary.go
```diff
@@ -531,7 +531,7 @@ func statusMessage(cols int, rows []*statepb.Row, recent int) string {
 const noGreens = "no recent greens"
 
 // latestGreen finds the ID for the most recent column with all passing rows.
-func latestGreen(grid *statepb.Grid) string {
+func latestGreen(grid *statepb.Grid, firstExtraColumn bool) string {
```
michelle192837: Parameter name is still confusing to me, maybe something like 'useFirstExtra'?

fejta: Fixed

fejta:
Now with passing unit tests??

PR #102, opened 2020-02-29
Handle scenario where grid is not found

fejta:

Use [error wrapping](https://blog.golang.org/go1.13-errors) to detect when the grid does not exist. Return a skeleton summary in this scenario.

Cover with unit tests


fejta:
passing tests now

PR #103, opened 2020-03-03
Continuing using gcs_prefix externally

fejta:
ref https://github.com/GoogleCloudPlatform/testgrid/pull/87

This is preventing me from updating testgrid in kubernetes/test-infra


pb/config/config.proto
```diff
@@ -58,10 +58,8 @@ message TestGroup {
   // Name of this TestGroup, for mapping dashboard tabs to tests.
   string name = 1;
 
-  // Path to the test result location, including protocol.
-  // The location must be accessible from the Update Server, and not local.
-  // GCS is usually used, making this field "gs://<bucket>"
-  string query = 2;
+  // Path to the test result stored in gcs
```
chases2: Please maintain the comment that protocol must be included.

I foresee a future where one tool is expecting this to be `gs://location` and another is `location`, and it's not clear which should be correct.

fejta: Done

fejta: The reason the original docs did not have this is because the protocol must be EXCLUDED:

https://github.com/kubernetes/test-infra/blob/9d2ed37f81a7a85216fa28845717beba19dcacef/config/testgrids/config.yaml#L10

PR #104, opened 2020-03-05
Validate things are balanced and not empty.

fejta:
Validate current incorrect column header behavior (will correct tomorrow).

ref https://github.com/kubernetes/test-infra/issues/16625

### DIFF
--- a/config/yamlcfg/yaml2proto_test.go
+++ b/config/yamlcfg/yaml2proto_test.go
@@ -256,6 +256,45 @@ test_groups:
 - name: testgroup_1
 `),
 		},
+		{
+			name: "column headers work poorly TODO(fejta): fix",
+			input: config.Configuration{
+				TestGroups: []*config.TestGroup{
+					{
+						Name: "test_group",
+						ColumnHeader: []*config.TestGroup_ColumnHeader{
+							{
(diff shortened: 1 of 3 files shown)


config/yamlcfg/yaml2proto_test.go
```diff
@@ -254,6 +254,45 @@ func Test_MarshalYAML(t *testing.T) {
   name: dashboard_1
 test_groups:
 - name: testgroup_1
+`),
+		},
+		{
+			name: "column headers work poorly TODO(fejta): fix",
```
chases2: Can you expand a bit here, perhaps in a comment? What is the TODO here?

config/yamlcfg/yaml2proto_test.go
```diff
@@ -254,6 +254,45 @@ func Test_MarshalYAML(t *testing.T) {
...
+			},
+			expected: []byte(`dashboards:
+- dashboard_tab:
+  - name: tab
+    test_group_name: test_group
+  name: dash
+test_groups:
+- column_header:
+  - ColumnHeaderSource:
+      ConfigurationValue: sad
```
fejta: This is... not correct

config/yamlcfg/yaml2proto_test.go
fejta: PR forthcoming

PR #105, opened 2020-03-05
Remove oneof for column headers

fejta:
Instead enforce that only one field is set ourselves. 

This allows [our yaml documents](https://github.com/kubernetes/test-infra/blob/e8ef9758b3abda6ffcb89385ae5e0e690615d66a/config/testgrids/default.yaml#L10-L12) to continue parsing this correctly:
```yaml
  column_header:
    - configuration_value: Commit # Shows the commit number on column header
    - configuration_value: infra-commit
```

/assign @chases2 @michelle192837


config/config.go
```diff
@@ -169,9 +169,12 @@ func validateReferencesExist(c configpb.Configuration) error {
 		tgNames[tg.Name] = true
 
 		for idx, header := range tg.ColumnHeader {
-			if header.GetConfigurationValue() == "" {
+			if cv, p, l := header.ConfigurationValue, header.Property, header.Label; cv == "" && p == "" && l == "" {
 				mErr = multierror.Append(mErr, fmt.Errorf("Colum Header %d in group %s is empty", idx, tg.Name))
+			} else if cv != "" && (p != "" || l != "") || p != "" && (cv != "" || l != "") {
+				mErr = multierror.Append(mErr, fmt.Errorf("Colum Header %d in group %s must only set one value, got configuration_value: %q, property: %q, label: %q", idx, tg.Name, cv, p, l))
```
michelle192837: nit: Column

fejta: Done

fejta:
It was not previously a oneof and must not be a oneof for yaml to unmarhsal it correctly.

From the proto perspective, the following two messages represent equivalent bytestreams (assuming only one field is set):

```
  message ColumnHeader {
    oneof column_header_source {
      string label = 1;
      string property = 2;
      string configuration_value = 3;
    }
}

message ColumnHeader {
    string label = 1;
    string property = 2;
    string configuration_value = 3;
  }
```

However these generate two different json and yaml documents. So the previous change to switch it from the latter to the former breaks json/yaml documents, which we obviously depend on (hence why upgrading testgrid in test-infra caused it to upload junk.

fejta:
Since k8s has historically only exposed the one configuration_value option (the other two fields have to do with resultstore which it doesn't integrate with yet) it was not necessary to validate that we only set at most 1 of the 1 possible options. Now that we have 3 it is good to add this validation.

michelle192837:
Makes sense, thanks for the explanation. ^^

PR #106, opened 2020-03-05
Ensure yaml config parses correctly.

fejta:


config/yamlcfg/yaml2proto_test.go
```diff
@@ -256,6 +256,33 @@ test_groups:
 - name: testgroup_1
 `),
 		},
+		{
```
michelle192837: One more case for rejecting empty column headers as well?

fejta: Done

PR #108, opened 2020-03-09
Advanced configuration option validation.

michelle192837:
- Validate that names don't have forbidden prefixes (e.g. can't start with the prefixes of special files we use in TestGrid)
- Validate TestGroup and DashboardTab-specific options

/assign fejta@, chases2@

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -18,6 +18,7 @@ package config
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -52,6 +53,7 @@ func (e DuplicateNameError) Error() string {
 	return fmt.Sprintf("found duplicate name after normalizing: (%s) %s", e.Entity, e.Name)
 }
 
+// MissingEntityError is an error that includes the missing entity.
 type MissingEntityError struct {
 	Name   string
 	Entity string
@@ -61,6 +63,7 @@ func (e MissingEntityError) Error() string {
 	return fmt.Sprintf("could not find the referenced (%s) %s", e.Entity, e.Name)
 }
 
+// ConfigError is an error for invalid configuration that includes what entity errored.
 type ConfigError struct {
 	Name    string
 	Entity  string
@@ -81,7 +84,7 @@ func normalize(s string) string {
 
 // validateUnique checks that a list has no duplicate normalized entries.
 func validateUnique(items []string, entity string) error {
-	mErr := &multierror.Error{}
+	var mErr error
 	set := map[string]bool{}
 	for _, item := range items {
 		s := normalize(item)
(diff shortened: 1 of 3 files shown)


config/config.go
```diff
@@ -225,6 +207,234 @@ func validateReferencesExist(c configpb.Configuration) error {
 	return mErr.ErrorOrNil()
 }
 
+// validateName validates an entity name is non-empty and contains no prefix that overlaps with a
+// TestGrid file prefix, post-normalization.
+func validateName(s string, entity string) error {
+	name := normalize(s)
+	if name == "" {
+		return &ConfigError{s, entity, "Normalized name can't be empty."}
```
fejta: a) How are we using these strongly typed errors?
b) Why do we need to return a reference?

config/config.go
```diff
@@ -225,6 +207,234 @@ func validateReferencesExist(c configpb.Configuration) error {
...
+		if err != nil {
+			mErr = multierror.Append(mErr, &ConfigError{dt.Name, "DashboardTab", err.Error()})
+		}
+	}
+
+	return mErr.ErrorOrNil()
+}
+
+func validateEntityConfigs(c configpb.Configuration) error {
+	mErr := &multierror.Error{}
```
fejta: `var mErr error`

https://github.com/hashicorp/go-multierror#usage

config/config.go
```diff
@@ -225,6 +207,234 @@ func validateReferencesExist(c configpb.Configuration) error {
...
+
+	// At the moment, don't need to further validate Dashboards or DashboardGroups.
+	for _, tg := range c.TestGroups {
+		err := validateTestGroup(tg)
+		if err != nil {
+			mErr = multierror.Append(mErr, err)
```
fejta: `mErr = multierror.Append(mErr, err)`

config/yamlcfg/yaml2proto_test.go
```diff
@@ -499,7 +521,10 @@ test_groups:
 			input: config.Configuration{
 				TestGroups: []*config.TestGroup{
 					{
-						Name: "test_group",
+						Name:             "test_group",
+						GcsPrefix:        "fake path",
```
fejta: FYI: I'm not super wild about having so many fields unrelated to "basic dashboard values work" as it makes these cases harder to understand what we're actually trying to test (not GcsPrefix for sure). I don't have a great solution though off-hand

config/config.go
michelle192837: Sorry, can't tell from this comment, what's the suggestion? (Just append without checking err != nil?, or something else?)

config/config.go
michelle192837: Most of what I want is to easily display all the errors present in the config to the end user when validating config, and make it clear what parts of the config need to be fixed (i.e. what type + name errored). I'm not sure if these kinds of errors are best or if I should just switch to structured logging at this point?

config/yamlcfg/yaml2proto_test.go
michelle192837: Yeeeeeah, I agree. I was thinking of breaking out a default TestGroup to add to the tests, but I'm not sure if the tradeoff on explicitness is worth it here.

config/config.go
michelle192837: Has to be a multierror error for ErrorOrNil, but changed.

config/yamlcfg/yaml2proto_test.go
michelle192837: Changed, might be better?

config/config.go
fejta: I don't see why all this code can't just be the more idiomatic:
```golang
var merr error
if err := foo(); err != nil {
  merr = multierror.Append(merr, err)
}
if err := bar(); err != nil {
  merr = multierror.Append(merr, err)
}
// etc
return merr
```

michelle192837: Ah, misunderstood. Done!

config/config.go
michelle192837: Think this is addressed now.

config/config.go
michelle192837: Refactored a bit to create ConfigErrors in fewer places.

PR #110, opened 2020-03-17
Config proto should be passed by reference.

michelle192837:


chases2:
The change seems fine, but what's the reasoning for this change? In Go, moving a constant struct from value to reference [isn't universally faster](https://medium.com/@vCabbage/go-are-pointers-a-performance-optimization-a95840d3ef85).

/lgtm
/hold

config/config.go
```diff
@@ -155,15 +155,15 @@ func validateAllUnique(c configpb.Configuration) error {
 	return mErr
 }
 
-func validateReferencesExist(c configpb.Configuration) error {
+func validateReferencesExist(c *configpb.Configuration) error {
```
chases2: If these function arguments are going to change to pointers, they now need null checks.

```
if c == null {
  return fmt.Errorf("descriptive message")
}
```

Similarly for other functions.

michelle192837:
The change is because of apparent potential issues with shallow-copying proto messages, not performance.

config/config.go
michelle192837: Fiiiiiiixed I think?

michelle192837: In this PR: Michelle learned the importance of using the dang proto methods and not just accessing things willy-nilly.

PR #111, opened 2020-03-19
Adding Search function to the resultstore client.

tony-yang:
Adding Search function to upstream TestGrid resultstore. Adding context as an argument to allow others to call the function and pass in their own context.

### DIFF
--- a/resultstore/client_test.go
+++ b/resultstore/client_test.go
@@ -1,0 +1,79 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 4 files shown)


resultstore/client_test.go
```diff
@@ -0,0 +1,77 @@
...
+							StartTime: now,
+							Duration:  duration,
+						},
+					},
+				},
+			},
+			expected: []*Invocation{
+				{
+					Name:     "invocations/fakeid-12345",
+					Duration: time.Duration(duration.Seconds)*time.Second + time.Duration(duration.Nanos)*time.Nanosecond,
```
chases2: I'm not sure what's going on here. It looks like you're counting the `duration` variable twice?

To clarify, consider either commenting or changing the `duration` variable name; it's tough to understand with the `Duration` key and the `Duration` library.

tony-yang: Thank you. Appreciate the feedback. Created functions to do the conversion with clear name.

PR #113, opened 2020-03-25
Change gcs_prefix back to query for consistency.

michelle192837:
This will be followed up with a PR in test-infra to change uses of gcs_prefix to query.

Needed so we can validate all types of config protos.


fejta:
I'm not a fan. GCSPrefix is the right name.

michelle192837:
It doesn't work for the internal instance, and it's going to require extensive patching to be able to use this internally with the field change. I'd really like to get us using the same config.proto internally and externally.

fejta:
We can use a different name internally

chases2:
I'd like to do a bit of research before weighing in here.
/hold

PR #118, opened 2020-04-13
Add README for the resultstore client

tony-yang:
Adding instruction and code snippet on invocation search.


tony-yang:
> Hmm, no site and end user can visit for this?

Unfortunately, search is API only, at least for now. Once they have the exact invocation name, they can plug that into source.cloud.google.com/results/invocations/<id> to see individual invocation, assuming if they have the permission to access that GCP project.

PR #120, opened 2020-05-14
Migrate the config.md to GoogleCloudPlatform/testgrid and update

MushuEE:
The [old config docs](https://github.com/kubernetes/test-infra/blob/master/testgrid/config.md) had some dead links and unresolved links due to the migration. This PR creates a new config.md for GCP/testgrid and links back to k8s/test-infra where needed. It also removes some elements that only exist from internal. 

Ex. <test-url> does not exist in the OSS context and was removed. 
Ex. <config.proto> paths changed

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ If you want to modify TestGrid beyond adding new tests or dashboards, see
 ## Configuration
 
 If you need to add a new test that you want TestGrid to display, or otherwise change what is shown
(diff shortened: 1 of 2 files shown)


MushuEE:
@michelle192837 not sure what to do about CLA. I signed it and it shows I have one?

MushuEE:
Closing and making new with email updated for CLA

PR #121, opened 2020-05-14
Migrate config.md from old k8s/testinfra to CGP/testgrid

MushuEE:
The old config docs had some dead links and unresolved links due to the migration. This PR creates a new config.md for GCP/testgrid and links back to k8s/test-infra where needed. It also removes some elements that only exist from internal.

Ex. does not exist in the OSS context and was removed.
Ex. <config.proto> paths changed


michelle192837:
Yup, this looks good! And yeah, agreed it should be removed from kubernetes/test-infra after this. ^^

/hold cancel

fejta:
/ok-to-test
thanks for fixing this btw!

PR #122, opened 2020-05-15
Update broken link format in config.md

MushuEE:
small fix on link

*I don't have a good way to test this. I would like to build a quick staging ENV using docker compose or something with mocked data to quickly test changes if possible


MushuEE:
I used [this](https://yaml-multiline.info/) as reference

chases2:
Hello, and thanks for your PR!

Are you trying to change the template for a particular live TestGrid? The YAML file here was forked from [kubernetes/test-infra](https://github.com/kubernetes/test-infra) as an example configuration file. It isn't live anywhere, and we don't have a test to ensure the example code is valid.

michelle192837:
My bad on that I think, I completely missed which file was being updated. ><

michelle192837:
Apologies on missing that before. I believe https://github.com/kubernetes/test-infra/blob/master/config/testgrids/default.yaml#L24 is what you want to update to update the default for everyone on testgrid.k8s.io.

MushuEE:
Ah ok I can revert the default template changes and just leave the link update!

michelle192837:
Sounds good!

PR #123, opened 2020-05-21
Add Column failure collection for use in 2nd stage processing (Part 1)

MushuEE:
This is part of an effort involving adding some new metrics for the failure ratios of columns. I changed the standard bool tracking of row pass rates to track the number of passes and fails. Then calculate the ratio of failing cells to total cells. 

* I also treated PASS_WTH_SKIPS and PASS_WITH_ERRORS as passing and added them into the unit tests. 

Q: Do we want running to be treated as failure still?

I also pulled the status string generation out and gave it it's own test.

Talking with fejta it seem there was a push to avoid adding any new deps and modules. Being able to use assert was nice because it allowed me to create a variadic shim to easily test functions with multiple different return types. If there is a way to do this easily without assert + shim I am all ears. 

Only about 2 months into playing with Golang and not aware of the design patterns so feel free to bring this PR to Gulag.

### DIFF
--- a/pkg/summarizer/summary.go
+++ b/pkg/summarizer/summary.go
@@ -246,7 +246,7 @@ func updateTab(ctx context.Context, tab *configpb.DashboardTab, findGroup groupF
 	latest, latestSeconds := latestRun(grid.Columns)
 	alert := staleAlert(mod, latest, staleHours(tab))
 	failures := failingTestSummaries(grid.Rows)
-	//TODO(gmccloskey): update DashboardTabSummary to inclused failure ratio in PB
+	//TODO(gmccloskey): update DashboardTabSummary to include failure ratio in PB
 	passingCols, completedCols, passingCells, filledCells, _ := gridMetrics(len(grid.Columns), grid.Rows, recent)
 	return &summarypb.DashboardTabSummary{
 		DashboardTabName:     tab.Name,
@@ -488,7 +488,7 @@ func overallStatus(grid *statepb.Grid, recent int, stale string, alerts []*summa
 	return summarypb.DashboardTabSummary_UNKNOWN
 }
 
-// Culminate set of metrics related to a section of the Grid that can be consumed by
+// Culminate set of metrics related to a section of the Grid
 func gridMetrics(cols int, rows []*statepb.Row, recent int) (int, int, int, int, []float32) {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
@@ -500,12 +500,16 @@ func gridMetrics(cols int, rows []*statepb.Row, recent int) (int, int, int, int,
 	var passingCols int
 	var completedCols int
 
-	for idx := 0; idx < cols && idx < recent; idx++ {
+	for idx := 0; idx < cols; idx++ {
+		if idx >= recent {
+			break
+		}
 		var passes int
 		var failures int
 		for _, ch := range results {
 			// TODO(fejta): fail old running cols
 			switch coalesceResult(<-ch, result.IgnoreRunning) {
+			//TODO(michelle192837): Create utility to standardize pass/fail boundaries
 			case statepb.Row_PASS, statepb.Row_PASS_WITH_ERRORS, statepb.Row_PASS_WITH_SKIPS:
 				passes++
 				passingCells++
--- a/pkg/summarizer/summary.go
+++ b/pkg/summarizer/summary.go
@@ -246,6 +246,7 @@ func updateTab(ctx context.Context, tab *configpb.DashboardTab, findGroup groupF
 	latest, latestSeconds := latestRun(grid.Columns)
 	alert := staleAlert(mod, latest, staleHours(tab))
 	failures := failingTestSummaries(grid.Rows)
+	//TODO(gmccloskey): update DashboardTabSummary to inclused failure ratio in PB
 	passingCols, completedCols, passingCells, filledCells, _ := gridMetrics(len(grid.Columns), grid.Rows, recent)
 	return &summarypb.DashboardTabSummary{
 		DashboardTabName:     tab.Name,
@@ -499,7 +500,7 @@ func gridMetrics(cols int, rows []*statepb.Row, recent int) (int, int, int, int,
 	var passingCols int
 	var completedCols int
 
-	for idx := 0; idx < cols && idx <= recent; idx++ {
+	for idx := 0; idx < cols && idx < recent; idx++ {
 		var passes int
 		var failures int
 		for _, ch := range results {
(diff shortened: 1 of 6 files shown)


pkg/summarizer/summary.go
```diff
@@ -246,14 +246,16 @@ func updateTab(ctx context.Context, tab *configpb.DashboardTab, findGroup groupF
 	latest, latestSeconds := latestRun(grid.Columns)
 	alert := staleAlert(mod, latest, staleHours(tab))
 	failures := failingTestSummaries(grid.Rows)
+	//TODO(gmccloskey): update DashboardTabSummary to inclused failure ratio in PB
```
michelle192837, line 249: nit: include?

pkg/summarizer/summary.go
```diff
@@ -486,53 +488,61 @@ func overallStatus(grid *statepb.Grid, recent int, stale string, alerts []*summa
...
 
-func fmtStatus(passCols, cols, passCells, cells int) string {
-	colCent := 100 * float64(passCols) / float64(cols)
-	cellCent := 100 * float64(passCells) / float64(cells)
-	return fmt.Sprintf("%d of %d (%.1f%%) recent columns passed (%d of %d or %.1f%% cells)", passCols, cols, colCent, passCells, cells, cellCent)
-}
-
-func statusMessage(cols int, rows []*statepb.Row, recent int) string {
-	//  2483 of 115784 tests (2.1%) and 163 of 164 runs (99.4%) failed in the past 7 days
+// Culminate set of metrics related to a section of the Grid that can be consumed by
```
michelle192837, line 491: Incomplete comment?

pkg/summarizer/summary.go
```diff
@@ -486,53 +488,61 @@ func overallStatus(grid *statepb.Grid, recent int, stale string, alerts []*summa
...
 	var passingCols int
 	var completedCols int
-	for idx := 0; idx < cols; idx++ {
-		if idx >= recent {
-			break
-		}
-		var passes bool
-		var failures bool
+
+	for idx := 0; idx < cols && idx < recent; idx++ {
```
michelle192837, line 503: nit: I think the explicit condition before is a bit more readable (if idx >= recent { break }).

pkg/summarizer/summary.go
```diff
@@ -486,53 +488,61 @@ func overallStatus(grid *statepb.Grid, recent int, stale string, alerts []*summa
...
+		var passes int
+		var failures int
 		for _, ch := range results {
 			// TODO(fejta): fail old running cols
 			switch coalesceResult(<-ch, result.IgnoreRunning) {
-			case statepb.Row_PASS:
-				if !failures {
-					passes = true
-				}
+			case statepb.Row_PASS, statepb.Row_PASS_WITH_ERRORS, statepb.Row_PASS_WITH_SKIPS:
```
michelle192837, line 509: Can you add a TODO (feel free to assign to me) to make 'what's a passing status' and 'what's a failing status' a common utility function?

We should probably open-source status.go to make sure everyone's consistent about these kinds of concepts, since it gets tricky fast...

pkg/summarizer/summary.go
```diff
@@ -486,53 +488,61 @@ func overallStatus(grid *statepb.Grid, recent int, stale string, alerts []*summa
...
+				passes++
 				passingCells++
 				filledCells++
 			case statepb.Row_NO_RESULT:
 				// noop
 			default:
```
michelle192837, line 515: ...for example, here, since there are technically some statuses that aren't failing OR passing, like flaky tests. (Doesn't need to be added this PR, but we should keep track of needing to add it).

pkg/summarizer/summary.go
```diff
@@ -486,53 +488,61 @@ func overallStatus(grid *statepb.Grid, recent int, stale string, alerts []*summa
...
-		if failures || passes {
+		if failures > 0 || passes > 0 {
 			completedCols++
 		}
-		if passes {
+		if failures == 0 && passes > 0 {
 			passingCols++
 		}
+
+		if passes+failures > 0 {
```
michelle192837, line 528: if completedCols > 0?

pkg/summarizer/summary_test.go
```diff
@@ -1267,35 +1277,181 @@ func TestStatusMessage(t *testing.T) {
...
-			name:   "not enough columns yet works just fine",
-			cols:   2,
+			name:   "half and half",
+			cols:   4,
 			recent: 50,
 			rows: []*statepb.Row{
 				{
-					Name:    "two passes",
-					Results: []int32{int32(statepb.Row_PASS), 2},
+					Name: "four pass",
```
michelle192837, line 1292: nit: keep 'passes' instead of 'pass'?

pkg/summarizer/summary_test.go
```diff
@@ -1267,35 +1277,181 @@ func TestStatusMessage(t *testing.T) {
...
-			passingCells: 6,
-			filledCells:  7,
+			passingCols:       2, // pass, fail, pass
+			filledCols:        3,
+			passingCells:      6,
+			filledCells:       7,
+			expectedFailRatio: []float32{0, 0.33333334, 0},
 		},
 		{
-			name:   "not enough columns yet works just fine",
```
michelle192837, line 1276: Would keep this test case name (I think it's more descriptive of the point of this test, which is that we can summarize even when we don't have all the 'recent' columns filled).

pkg/summarizer/summary_test.go
```diff
@@ -1267,35 +1277,181 @@ func TestStatusMessage(t *testing.T) {
...
+	cases := []struct {
+		name             string
+		passingCols      int
+		completedCols    int
+		passingCells     int
+		filledCells      int
+		expectedOverride string
+	}{
+		{
+			name:             "no filledCells",
```
michelle192837, line 1414: This is the same as "all no result" below I think?

pkg/summarizer/summary_test.go
```diff
@@ -1105,21 +1106,24 @@ func TestOverallStatus(t *testing.T) {
 	}
 }
 
-func TestStatusMessage(t *testing.T) {
+func makeShim(v ...interface{}) []interface{} {
```
michelle192837, line 1109: Huh, not sure what my opinion is on makeShim; my intuition says this should be an explicit struct instead?

@fejta, opinion on this specifically?

pkg/summarizer/summary.go
MushuEE: I agree thanks, I am happy to try to do it as well!

pkg/summarizer/summary.go
MushuEE: Yeah was not sure how to bin those, we could add metrics for that bin as well if we wanted and track a tuple of status... though not sure how much granularity we care about.

pkg/summarizer/summary_test.go
MushuEE: Good point, because it is not being used generally I can just wrap with a struct.

pkg/summarizer/summary.go
michelle192837: I think we care less about 'granularity' than 'the concepts of passing and failing are consistent everywhere'. I don't think you have to do it this PR, but if you're interested we can sync some on doing that in a follow-up PR. ^^

pkg/summarizer/summary.go
MushuEE: The only reason I would hesitate is if we want to be explicit about what we are trying to avoid (div by 0 error) when calculating

MushuEE: I lied, `completedCols` spans columns and I if the next column has only `NO_RESULT` I do not want to allow us to div by 0

pkg/summarizer/summary_test.go
MushuEE: Thanks, recreating that test and splitting from 50/50

pkg/summarizer/summary_test.go
MushuEE: done

pkg/summarizer/summary_test.go
MushuEE: true... I was trying to be explicit and not thinking about the fact that it is just the base value for int

michelle192837:
Mmk, I think this should be good. Thanks Grant! ^^

PR #130, opened 2020-06-17
Add per-Tab BROKEN threshold Config option

MushuEE:
This will be used to enable per-tab control of what threshold levels will yield a BROKEN tab.


fejta:
I'd suggest creating an issue and linking the various PRs related here.

ref https://github.com/GoogleCloudPlatform/testgrid/pull/129


Thoughts on waiting until something uses these to check them in?

michelle192837:
I think we can wait until something uses them, yeah. ^^ Should be safe to include with a change like that.

The other thing is that until the config.proto in this repo is source-of-truth, we need to replicate this change internally first. (That's what the `hold` earlier was.)

michelle192837:
Ah, and linking an issue is just mentioning `#<issue num>` in the comments, like Erick did above with 129.

fejta:
If you paste a link to a github issue X (even from another org) in github issue Y it will create a reference between them (aka you can click on the link to Y in issue X and github will automatically create a reference in Y to note that X refers to it -- for example: https://github.com/GoogleCloudPlatform/testgrid/issues/26#ref-pullrequest-504317659)

MushuEE:
Now used for #133 however I think we need #132 to go in first and I need to make this change in sync with internal

MushuEE:
Pulled in with #132

PR #132, opened 2020-06-19
Updating config.proto.

michelle192837:
Also adding a warning note about updating config/state.proto before
updating the other version of those files.

I think since this option is not used, changing its type
and name is safe, but I'm not 100% sure on that.

(The replaced field will need to be added first in the other version of
config.proto, and then added back here.)


michelle192837:
/assign @fejta @chases2 
/cc @MushuEE

(Sorry about this! Can also try and update this by using new field numbers for the potentially-conflicting field 9 in AutoBugOptions)

pb/config/config.proto
```diff
@@ -2,6 +2,8 @@ syntax = "proto3";
 
 // Protocol buffer for configuring testgrid.k8s.io
 
+// NOTE: Do NOT update this until you have updated the other config.proto!
```
MushuEE, line 5: Is there anything we can put to decrease the ambiguity of "the other"?

michelle192837: Changed to 'internal' to be clearer. ^^

michelle192837:
/hold cancel

Thanks all!

PR #133, opened 2020-06-22
Logic to set a Tab Summary overall Status to BROKEN

MushuEE:
If one or more recent columns have a test cycle that has a failure ratio beyond the configured tab threshold, the over all state of the tab will be BROKEN.

If the failure ratio is unset or set to 0.0, it will be ignored and never enter Broken state


MushuEE:
/hold 
Need to make config changes first

pkg/summarizer/summary.go
```diff
@@ -527,14 +529,14 @@ func gridMetrics(cols int, rows []*statepb.Row, recent int) (int, int, int, int,
...
-		} else {
-			columnFailureRatios = append(columnFailureRatios, 0.0)
+		if passes+failures > 0 && brokenThreshold > 0 {
+			if float32(failures)/float32(passes+failures) > brokenThreshold {
+				brokenState = true
+			}
 		}
 	}
 
-	return passingCols, completedCols, passingCells, filledCells, columnFailureRatios
```
MushuEE, line 537: Revert config column failure ratio to reserved, it will no longer be used

MushuEE: done

Issue #137, opened 2020-06-26
Boiler plate specifies "The Kubernetes Authors", should be "Testgrid"


MushuEE:
The templates within Hack look like they were moved over from the k8s project. Now that Testgrid has migrated to its own project. It should have updated Copyright.

PR #138, opened 2020-06-26
Update boilerplate template and file headers to use "The Testgrid Authors"

MushuEE:
Update all current headers and boilerplate templates to replace Kubernetes->Testgrid

#137 
/cc @itsazhuhere

### DIFF
--- a/hack/boilerplate/boilerplate.Dockerfile.txt
+++ b/hack/boilerplate/boilerplate.Dockerfile.txt
@@ -1,4 +1,4 @@
-# Copyright YEAR The Kubernetes Authors.
+# Copyright YEAR The Testgrid Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/hack/boilerplate/boilerplate.Dockerfile.txt
+++ b/hack/boilerplate/boilerplate.Dockerfile.txt
@@ -1,4 +1,4 @@
-# Copyright YEAR The Testgrid Authors.
+# Copyright YEAR The AUTHOR Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/hack/verify_boilerplate.py
+++ b/hack/verify_boilerplate.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 
-# Copyright 2015 The Kubernetes Authors.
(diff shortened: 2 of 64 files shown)


MushuEE:
/hold
@michelle192837 Some of these files were ported over from K8s project. @itsazhuhere mentioned that we could check for Kubetnetes || Testgrid so that Copywrite check would pass for both types. Should that be the route I take? Or move all to Testgrid like this PR is currently?

michelle192837:
We should check for both. We should not be changing the headers on existing files ("Kubernetes Authors" has to remain for files we forked from kubernetes/test-infra; it's only new files that are created from scratch in this repo that we can claim our own copyright on). So this repo reasonably can/should have a mix of different attributions. ^^

hack/verify_boilerplate.py
```diff
@@ -121,7 +123,14 @@ def file_passes(filename, refs, regexs):  # pylint: disable=too-many-locals
     # Replace all occurrences of the regex "2017|2016|2015|2014" with "YEAR"
     when = regexs["date"]
     for idx, datum in enumerate(data):
-        (data[idx], found) = when.subn('YEAR', datum)
+        (data[idx], found) = when.subn("YEAR", datum)
+        if found != 0:
+            break
+
+    # Replace all occurrences of the regex "Testgrid|Kubernetes" with "AUHTOR"
```
itsazhuhere, line 130: nit: this should be "AUTHOR"

hack/verify_boilerplate.py
```diff
@@ -24,6 +24,8 @@
 import re
 import sys
 
+AUTHORS = r"Testgrid|Kubernetes"
```
itsazhuhere, line 27: I'm pretty sure we want it properly capitalized as "TestGrid"

hack/boilerplate/boilerplate.Dockerfile.txt
```diff
@@ -1,4 +1,4 @@
-# Copyright YEAR The Kubernetes Authors.
+# Copyright YEAR The Testgrid Authors.
```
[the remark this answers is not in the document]
itsazhuhere: Oh woops I guess I didn't hit submit for this comment earlier ^.

michelle192837:
Huh, what's with all the proto changes?

michelle192837:
Oh wait, duh, comment changes.

Not sure why types.proto is also getting whitespace changes, though, but not a big deal.

PR #90, opened 2020-02-07
Update cloud.google.com/go dependency to use new storage package

saschagrunert:
Hey, we would like to update our golang dependencies in in https://github.com/kubernetes/release/pull/1067, but for that we would need to update the cloud.google.com/go package here which has been split into a dedicated `storage` package.

PR #141, opened 2020-07-07
Update testgrid's bazel and cloud.google.com/go dependencies

spiffxp:
I tried upgrading the cloud dependency but was unable to get bazel to work with it unless I upgraded to bazel 2.2.0.  For grins I tried bumping to 3.0.0 locally and had no problems.

Getting this to pass in CI will be more involved though. I was thinking about following the same dance we do for kubernetes/test-infra (ref: https://github.com/kubernetes/test-infra/tree/master/images#updating-test-infra-images) which would require making a testgrid-specific dual-bazel-version image and updating jobs accordingly.

I wanted to get eyes on this PR before going through that


spiffxp:
Supercedes https://github.com/GoogleCloudPlatform/testgrid/pull/90

spiffxp:
OK, looks like CI is happy with bazel 3.0.0 and the `cloud.google.com/go` -> `cloud.google.com/go/storage` change

fejta:
weird and great!

PR #142, opened 2020-07-08
Change channel usage logic to drain channels

itsazhuhere:
The current implementation, where it iterates through each column and grabs one output from each channel, causes summarizer to run much more slowly than if it were to consume the all of the output from one channel at a time. This change aims to fix that. 

It also adds logging of the timing for updateOnce and a way to track per row indices of messages (if a Row_Result is NO_RESULT then we don't increment the index, however for all other Row_Result's we do) for accuracy.

### DIFF
--- a/cmd/summarizer/main.go
+++ b/cmd/summarizer/main.go
@@ -84,6 +84,7 @@ func main() {
 		defer cancel()
 		return summarizer.Update(ctx, client, opt.config, opt.concurrency, opt.dashboard, opt.confirm)
 	}
+
 	if err := updateOnce(ctx); err != nil {
 		logrus.WithError(err).Error("Failed update")
 	}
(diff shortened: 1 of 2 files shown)


cmd/summarizer/main.go
```diff
@@ -84,10 +84,11 @@ func main() {
 		defer cancel()
 		return summarizer.Update(ctx, client, opt.config, opt.concurrency, opt.dashboard, opt.confirm)
 	}
-
+	start := time.Now()
 	if err := updateOnce(ctx); err != nil {
 		logrus.WithError(err).Error("Failed update")
 	}
+	logrus.Infof("updateOnce took %d ms to complete\n", int64(time.Since(start)))
```
fejta, line 91: time.Duration is the correct type here, not int64. Either way, ms is the wrong units for something which takes many minutes and sometimes more than an hour.

Look at https://golang.org/pkg/time/#Duration.Round

Also you can already calculate this information by looking at the deltas between the Sleeping log lines on L102.

itsazhuhere: Oh okay, thanks for the heads up. I guess since we already have this functionality then I'll just remove this. But I'll definitely keep that in mind for the future!

PR #150, opened 2020-07-15
Update prod updater to 20200714, turn off confirm.

fejta:
Also fix canary updater location


fejta:
Michelle you around to review this one?

PR #157, opened 2020-07-21
Prevent writing grid state to gs://k8s-testgrid, allow a subdir

fejta:


cmd/updater/main.go
```diff
@@ -57,8 +57,8 @@ func (o *options) validate() error {
 	if o.config.String() == "" {
 		return errors.New("empty --config")
 	}
-	if o.config.Bucket() == "k8s-testgrid" && o.config.Object() != "beta/config" && o.confirm { // TODO(fejta): remove
-		return fmt.Errorf("--config=%s cannot write to gs://k8s-testgrid/config", o.config)
+	if o.config.Bucket() == "k8s-testgrid" && o.gridPrefix != "" && o.confirm {
+		return fmt.Errorf("--config=%s: cannot write grid state to gs://k8s-testgrid", o.config)
```
michelle192837, line 61: This reads a bit confusing, can the error indicate that you need grid-prefix or that you're avoiding collision/the top-level directory?

fejta: Yes, any of the following are fine:
* Leave `--confirm=false` so we don't write anything
* Set `--config=some-other-bucket` so we aren't using our prod
* Leave `--grid-prefix=grid` (or anything other than `""`) so we write grid state somewhere different

This is to ensure the python and golang updates to not compete with each other.

michelle192837: That makes sense! My nit was that the error message looks like writing to the k8s-testgrid _bucket_ isn't possible, not that writing to gs://k8s-testgrid specifically isn't. (i.e. if I were looking at the error, I'd assume I can't provide k8s-testgrid as an arg, not that I also need to provide grid-prefix.)

Not a huge deal though.

PR #158, opened 2020-07-21
Add fields and logic to summarizer to also record previous flakiness values 

itsazhuhere:


pb/summary/summary.proto
```diff
@@ -77,6 +78,9 @@ message TestInfo {
   // The flakiness of the test, measured out of 100
   float flakiness = 8;
 
+  // The flakiness of the test from the previous interval
+  float previous_flakiness = 10;
```
michelle192837, line 82: Consider having this be repeated, if we'd ever be interested in looking at or comparing multiple previous flakiness values? (Can have it limited in the code so it doesn't grow unbounded, e.g. for now we can make it repeated but in the code ensure we only assign 1 value to it ever).

pb/summary/summary.proto
```diff
@@ -110,6 +114,8 @@ message HealthinessInfo {
   // The flakiness out of 100 (think percentage but drop the sign)
   float average_flakiness = 4;
 
+  // The average flakiness for the last interval
+  float previous_flakiness = 5;
```
michelle192837, line 118: Should this also be repeated?

pb/summary/summary.proto
itsazhuhere: Okay, done

pb/summary/summary.proto
itsazhuhere: Mm yeah, I agree. If TestInfo's should be repeated then it should be consistent.

pb/summary/summary.proto
```diff
@@ -110,6 +114,8 @@ message HealthinessInfo {
   // The flakiness out of 100 (think percentage but drop the sign)
   float average_flakiness = 4;
 
+  // The average flakiness for previous intervals
+  repeated float previous_flakiness = 5;
```
fejta, line 118: Why is this in two locations? Why does this need to be repeated? How will we decouple this value from the highly variable update cycle?

It seems like a timeseries database would be a better way to store and measure this sort of things.

michelle192837: The two locations are overall vs. per-test
And this is per-interval, so it shouldn't be dependent on exactly when the cycle updates, but on past time intervals.

Are you suggesting to remove the idea of a previous flakiness number altogether, or to make it non-repeated and store this in a database for retrieval?

michelle192837: (To clarify, my thought on this field is that this could have things like "current flakiness in last X days" and also "flakiness from X to 2X days ago", "flakiness from 2X to 3X days ago", etc, not necessarily things spanning longer than the history of the tab. (For longer historical data, timeseries database seems fine)

PR #160, opened 2020-07-23
Restore some state proto documentation

fejta:


pb/state/state.proto
```diff
@@ -90,20 +90,16 @@ message TestMetadata {
...
 
   // Name associated with the column (such as the run/invocation ID).No two
   // columns should have the same build_id and name. The name field allows the
   // display of multiple columns with the same build_id.
   string name = 2;
 
-  // Timestamp for the column, typically the earliest timestamp of any test
-  // result in the column.
-  double started = 3;
+  double started = 3; // Milliseconds since start of epoc (python time.time() * 1000)
```
michelle192837, line 100: nit: epoch?

fejta: Done

PR #161, opened 2020-07-23
Functions to reinflate compressed grid messages.

fejta:
Correct a documentation error on metric encoding.

This will soon be used to refactor the updater to only download the
latest results, rather than everything.

Sending just this as a way to break things up.

### DIFF
--- a/pkg/updater/inflate.go
+++ b/pkg/updater/inflate.go
@@ -1,0 +1,189 @@
+/*
+Copyright 2020 The TestGrid Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


pkg/updater/inflate.go
```diff
@@ -0,0 +1,188 @@
...
+// This includes both:
+// * Column state metadata and
+// * cell values for every row in this column
+type inflatedColumn struct {
+	column *state.Column
+	cells  map[string]cell
+}
+
+// cell holds row values for a given column
+type cell struct {
```
michelle192837, line 37: nit: `cell` sounds confusing since I typically think of a single square for that. (If Go's thing weren't called a `slice` I would suggest `slice` or `columnSlice` or similar).

pkg/updater/inflate_test.go
```diff
@@ -0,0 +1,601 @@
...
+		name     string
+		indices  []int32
+		values   []float64
+		expected []*float64
+	}{
+		{
+			name: "basically works",
+		},
+		{
+			name:    "documented example works",
```
michelle192837, line 504: I think "documented example" should be more descriptive here (and below), it's unclear where the documented examples are and I shouldn't need to cross-reference.

pkg/updater/inflate.go
fejta: Updated the doc, this is just a square (the inflatedColumns has a map[string]cell)

michelle192837:
Sweet, thanks! ^^

PR #163, opened 2020-07-23
Add more state proto Result fields and better pass/fail categorization

itsazhuhere:
Previously the external Row_Result fields did not contain categories such as CATEGORIZED_FAILURE, but their enum values were still passed in to Summarizer. This resulted in Coalesce defaulting those values to PASS. This update aims to at least correctly categorize failures; some categories such as UNKNOWN still default to PASS (but this matches the previous behavior).

Also adds a fix for an edge case and adjusts tests.


internal/result/results.go
```diff
@@ -29,13 +30,54 @@ const (
...
+}
+
 // Coalesce reduces the result to PASS, NO_RESULT, FAIL or FLAKY.
 func Coalesce(result state.Row_Result, ignoreRunning bool) state.Row_Result {
 	// TODO(fejta): other result types, not used by k8s testgrid
 	if result == state.Row_NO_RESULT || result == state.Row_RUNNING && ignoreRunning {
 		return state.Row_NO_RESULT
 	}
-	if result == state.Row_FAIL || result == state.Row_RUNNING {
+	if IsFailingResult(result) || result == state.Row_RUNNING {
```
fejta, line 80: @michelle192837 this deserves unit testing. 

Looks like CANCEL coalesces to passed?

fejta: https://github.com/GoogleCloudPlatform/testgrid/pull/266

PR #174, opened 2020-07-31
Add ReadColumns coverage.

fejta:
Also catch and resolve some bugs in the process:
* Need to get column header's configuration value.
* Distinguish between stop time and build download duration
* Truncate excessive results correctly
* Include the newest result that is too old.
* Fix a NPE when started/finished error.
* Avoid hang when multiple builds find a too old build concurrently


fejta:
Fixed the update-deps issue, now passing tests

Issue #184, opened 2020-08-05
Updating is too slow


fejta:
Looks like it will currently take hours to do an update. Need to figure out what is slowing us down and why.

/assign

fejta:
* race condition in readColumns (fix - #185)
* Spend too much time in ListBuilds -- just list recent builds, not all of them

PR #185, opened 2020-08-05
Fix race condition when multiple go-routines reach an old build

fejta:
ref https://github.com/GoogleCloudPlatform/testgrid/issues/184


fejta:
(rebased to pick up your recent delinting PR Michelle)

PR #189, opened 2020-08-05
Refactor Objects() to include an optional start offset

fejta:
This dramatically increases `ListBuilds()` performance for jobs we have recently compiled. For example:

* `--test-group=ci-kubernetes-e2e-gce-cos-k8sbeta-slow` takes 200ms to list 5 builds rather than 17500ms to list 2048 builds

Requires bumping to a prerelease version of cloud.google.com/go/storage as the lastest v1.10.0 release does not yet include https://github.com/googleapis/google-cloud-go/pull/2563


fejta:
ref #184

Issue #194, opened 2020-08-06
Updater consuming too much memory


fejta:
Lots of evictions, probably need to more precisely limit concurrency
/assign

fejta:
Things have improved with the new e2-highmem-4 instances and the above mitigations.

PR #202, opened 2020-08-07
Run the garbage collector after updating each group.

fejta:
This should reduce the amount of time the system is holding on to dead memory.


fejta:
ref https://github.com/GoogleCloudPlatform/testgrid/issues/194

PR #209, opened 2020-08-15
Record all issues linked to a tab in summary.

michelle192837:
This helps us in two ways:

1) We know which issues are actually linked in a table, as opposed to issues that MIGHT be linked in issue state (with the bonus that it gives us the issues after filtering! :D )
2) We can display a list of all linked issues in the tab summaries on the Summary tab.


pkg/summarizer/summary.go
```diff
@@ -522,6 +523,20 @@ func overallStatus(grid *statepb.Grid, recent int, stale string, brokenState boo
 	return summarypb.DashboardTabSummary_UNKNOWN
 }
 
+func allLinkedIssues(rows []*statepb.Row) []string {
+	issueMap := make(map[string]bool)
```
MushuEE, line 527: Is `IssueSet` a better name? or `uniqueIssues` 

https://dave.cheney.net/practical-go/presentations/qcon-china.html#_identifier_length

in his guidelines he recommended not including type in the name.

pkg/summarizer/summary_test.go
```diff
@@ -452,6 +455,70 @@ func TestRecentColumns(t *testing.T) {
...
+				},
+				{
+					Name:    "test-3",
+					Results: []int32{int32(statepb.Row_PASS), 10},
+				},
+			},
+			want: []string{},
+		},
+		{
+			name: "multiple linked issues",
```
MushuEE, line 488: Maybe also one for dupe issues?
```
		{
			name: "duplicate linked issues",
			rows: []*statepb.Row{
				{
					Name:    "test-1",
					Results: []int32{int32(statepb.Row_PASS), 10},
				},
				{
					Name:    "test-2",
					Results: []int32{int32(statepb.Row_PASS), 10},
                                   BugId:   []string{"1", "2"},
				},
				{
					Name:    "test-3",
					Results: []int32{int32(statepb.Row_PASS), 10},
                                   BugId:   []string{"2", "3"},
				},
			},
			want: []string{"1","2","3"},
		},
```

pkg/summarizer/summary.go
michelle192837: I think issueSet helps describe what it's actually doing so using that!

michelle192837:
Good suggestions, thanks! Done ^^

Issue #186, opened 2020-08-05
Add consistent idea of 'failing' and 'passing' statuses


michelle192837:
Right now we special-case these, but we should have a common method to use within TestGrid code that standardizes what counts as 'failing' vs. 'passing' statuses, since there are multiple.

(We have utility methods for this internally, I basically just want to make sure we use them here too).

/assign

michelle192837:
They do, though in a bit of a weird state (we have them in state.proto instead of using status.proto's enum, that should be an easy fix). We can't fully centralize this logic since the updater is the one calculating it, and any version of the updater needs to do this, but I agree we should have just one version in Go.

michelle192837:
I think we can call this good for now; Go code is now using the same code to compare statuses, and get passing/failing ones.

PR #224, opened 2020-08-21
Write a new flakiness algorithm for the summarizer

RonWeber:
By counting changes from pass->fail and ignoring long runs of failures, we can get a value for TestInfo.Flakiness that is more predictive of what tests are the most flaky.


michelle192837:
General question for folks; do you think it's better to replace the base analyzer here, or make it an option to pick which?

(Since GKE is the main use right now, I think either is plausible; either we just swap to a new analyzer based on the old one that takes status flips into account, or we let users pick which to use based on what's more useful).

pkg/summarizer/flakiness.go
```diff
@@ -102,9 +102,13 @@ func parseGrid(grid *statepb.Grid, startTime int, endTime int) []*common.GridMet
 	gridMetricsMap := make(map[string]*common.GridMetrics, 0)
 	gridRows := make(map[string]*statepb.Row)
 
+	// For each filtered test, status of non-infra-failure tests
+	relevantFilteredStatus := make(map[string][]statepb.Row_Result)
+
 	for i, row := range grid.Rows {
 		gridRows[row.Name] = grid.Rows[i]
 		gridMetricsMap[row.Name] = common.NewGridMetrics(row.Name)
+		relevantFilteredStatus[row.Name] = []statepb.Row_Result{}
```
michelle192837, line 111: nit: Maybe `rowStatuses` or similar?

pkg/summarizer/enhancedanalyzer/analyzer.go
```diff
@@ -0,0 +1,105 @@
...
+	averageFlakiness := float32(0)
+	for _, test := range hi.Tests {
+		test.Flakiness = calculateEnhancedFlakiness(relevantFilteredStatus[test.DisplayName])
+		averageFlakiness += test.Flakiness
+	}
+	hi.AverageFlakiness = averageFlakiness / float32(len(hi.Tests))
+	return hi
+}
+
+const IGNORE_FAILURES_IN_A_ROW = 3
```
michelle192837, line 54: Is this existing behavior? I know this is supposed to discount 'actual failures' from counting as flakes, but I also feel like this is kinda specific to GKE's use case? Status flips should be able to tell that kind of story on their own, _or_ we should use the configured failure threshold (or whether the row has an 'alert' on it) to determine.

pkg/summarizer/enhancedanalyzer/analyzer.go
```diff
@@ -0,0 +1,105 @@
...
+		if cf >= IGNORE_FAILURES_IN_A_ROW {
+			// Ignore the run of failures
+			i += cf
+			if i >= len(statuses) {
+				break
+			}
+		}
+		s := statuses[i]
+		considered++
+		if s == statepb.Row_PASS {
```
michelle192837, line 85: This set of statuses (PASS, FLAKY, other) won't be sufficient since multiple statuses can be 'passing' and multiple statuses can be 'failing', though it might be okay for particular dashboards. I'd either fix that, or note that in the comments here and add a TODO linked to https://github.com/GoogleCloudPlatform/testgrid/issues/186 to properly check what's passing vs. failing.

fejta:
```
ERROR: bazel rules out of date. Fix with:
  bazel run @io_k8s_repo_infra//hack:update-bazel
```

pkg/summarizer/enhancedanalyzer/analyzer.go
```diff
@@ -0,0 +1,105 @@
...
+
+// EnhancedAnalyzer implements functions that calculate flakiness as a ratio of failed tests to total tests
+type EnhancedAnalyzer struct {
+}
+
+// GetFlakiness returns a HealthinessInfo message
+func (ea *EnhancedAnalyzer) GetFlakiness(gridMetrics []*common.GridMetrics, relevantFilteredStatus map[string][]statepb.Row_Result, minRuns int, startDate int, endDate int, tab string) *summarypb.HealthinessInfo {
+	// Delegate to a BaseAnalyzer and change the flakiness scores.
+	// (And average flakiness)
+	na := baseanalyzer.BaseAnalyzer{}
```
fejta, line 43: `var na baseanalyzer.BaseAnalyzer` (you are not inializing anything)

(please apply this everywhere you declare a variable at its initial value (aka `:= float32(0)`, `:= 0`)

pkg/summarizer/baseanalyzer/analyzer.go
```diff
@@ -14,29 +14,28 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package naiveanalyzer
+package baseanalyzer
```
fejta, line 17: This one function does not need its own package. How about renaming it to analyzers and putting both of them in here?

pkg/summarizer/flakiness.go
```diff
@@ -38,15 +38,15 @@ var (
 )
 
 type flakinessAnalyzer interface {
-	GetFlakiness(gridMetrics []*common.GridMetrics, minRuns int, startDate int, endDate int, tab string) *summarypb.HealthinessInfo
+	GetFlakiness(gridMetrics []*common.GridMetrics, relevantFilteredStatus map[string][]statepb.Row_Result, minRuns int, startDate int, endDate int, tab string) *summarypb.HealthinessInfo
```
fejta, line 41: This interface seems like the wrong abstraction:
* the second instance of the interface already had to change the signature
* interfaces are already ignoring some of the arguments.

pkg/summarizer/enhancedanalyzer/analyzer.go
RonWeber: This is existing behavior in the GKE-specific autojailer.  It exists to handle a (depressingly common) case where a test is legitimately failing for most of the time period and flaky for the rest of it, and so looks healthy.

pkg/summarizer/baseanalyzer/analyzer.go
RonWeber: Done.

pkg/summarizer/flakiness.go
RonWeber: I changed the way we used the interface a little.

pkg/summarizer/flakiness.go
RonWeber: Done.

pkg/summarizer/enhancedanalyzer/analyzer.go
RonWeber: I changed this to be a specific set of values that exists only here.

A bit awkward, but at least it's not misleading.

pkg/summarizer/enhancedanalyzer/analyzer.go
RonWeber: Done.

pkg/summarizer/enhancedanalyzer/analyzer.go
itsazhuhere: Yeah, the statuses here come from results.Coalesce and other filtering in the new code. Whether the values from Coalesce are right is another story, but I think they should be pretty similar and if need be, Coalesce can be changed or we can use a different function that works similarly and covers everything correctly. I think for now something helpful for clarity would be to add a comment mentioning that these values have been filtered previously. Also this should be a switch case statement for consistency with other parts of the code.

pkg/summarizer/analyzers/flipanalyzer.go
```diff
@@ -0,0 +1,108 @@
...
+	// Delegate to a BaseAnalyzer and change the flakiness scores.
+	// (And average flakiness)
+	var ba BaseAnalyzer
+	healthinessInfo := ba.GetFlakiness(gridMetrics, minRuns, startDate, endDate, tab)
+	var averageFlakiness float32
+	for _, test := range healthinessInfo.Tests {
+		test.Flakiness = calculateFlipFlakiness(ea.RelevantStatus[test.DisplayName])
+		averageFlakiness += test.Flakiness
+	}
+	healthinessInfo.AverageFlakiness = averageFlakiness / float32(len(healthinessInfo.Tests))
```
itsazhuhere, line 53: This should have a check on the size of hi.Tests to prevent a divide by zero operation (which goes relatively undetected in Go).

RonWeber: Done.

michelle192837:
Thanks Ron! LGTM from me

pkg/summarizer/analyzers/flipanalyzer.go
```diff
@@ -0,0 +1,112 @@
...
+	StatusFail
+	StatusFlaky
+)
+
+// GetFlakiness returns a HealthinessInfo message
+func (ea *FlipAnalyzer) GetFlakiness(gridMetrics []*common.GridMetrics, minRuns int, startDate int, endDate int, tab string) *summarypb.HealthinessInfo {
+	// Delegate to a BaseAnalyzer and change the flakiness scores.
+	// (And average flakiness)
+	var ba BaseAnalyzer
+	healthinessInfo := ba.GetFlakiness(gridMetrics, minRuns, startDate, endDate, tab)
```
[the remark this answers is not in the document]
RonWeber: That confused me to, but it's a little out of scope for this PR.

fejta:
FYI leave a `/hold cancel` as the equivalent of clicking the submit button

PR #227, opened 2020-09-04
Use test_status.proto.

michelle192837:
Making this consistent with internal use of the enum. (Not actually sure how big a deal it is but I think it'll make consolidating 'this is status a failing status' logic easier).


chases2 (approved):
Collapsing these proto enums seems useful to me from a code health perspective.

Issue #36, opened 2019-10-17
Clean up config.proto options


michelle192837:
We should logically group together some of the options in config.proto (see AutoBugOptions). This will make it easier to parse what options are related to each other, and also make it easier to extend options in future without resorting to a lot of searching and sorting of proto fields.

PR #229, opened 2020-09-08
Deduplicate alerting settings in proto

chases2:
Addresses #36

Deduplicates alerting options to be contained in a Dashboard Tab instead of the underlying Test Group.
This gives an authoritative place where these setting can be stored or read from, and makes explicit an intent to migrate in this direction.


pb/config/config.proto
```diff
@@ -607,6 +604,14 @@ message DashboardTabAlertOptions {
 
   // Custom text to show for the debug link.
   string debug_message = 7;
+
+  // Wait time between emails. If unset or zero, an email will be sent only once
+  // it becomes a consistent failure, and not again until it succeeds.
+  // TestGrid does not pester about staleness
```
michelle192837, line 610: nit: Maybe clarify "Regardless of how this is set, TestGrid does not pester about staleness" ?

michelle192837: Also is this true?

michelle192837 (approved):
/hold

Unhold at will!

michelle192837:
Actually, updating that to:

/hold

Unhold once internal config.proto is updated.

pb/config/config.proto
chases2: I think this is true in the current implementation. Is it supposed to be different?

michelle192837: Nope, think that's fine, just wasn't sure. Carry on!

PR #230, opened 2020-09-09
Update test status severity.

michelle192837:


e-blackwelder:
I see that the statusSeverity is being sorted differently. Are the "enum" values being defined, used, or stored elsewhere? (In a previous job, I noticed an API's enum values being changed overtime, resulting in client-side bugs. I'm worried that something similar might be happening here as well.)

michelle192837:
Severity is specifically to avoid remapping enum values; we keep a separate map of 'this status should show up over this other status' to keep it separate from the defined enum in the proto.

michelle192837:
To be clear, the updated status severity sorting is intentional, this should make things more consistent between parts of TestGrid.

e-blackwelder (approved):

PR #234, opened 2020-09-13
Migrate to new sortorder import path

fvbommel:
See fvbommel/util#7

Note that I'm unfamiliar with Bazel, so please be extra critical of those changes.
I think I got it right, but don't be afraid to tell me if I screwed up :sweat_smile:.


michelle192837:
Sorry for the delay!

/ok-to-test

michelle192837:
So assuming tests pass, I think this swap should be fine, but it depends on how it integrates with other code. I may end up needing to revert this, but will let you know if so. ^^

michelle192837 (approved):

fvbommel:
> Sorry for the delay!

No problem. It's not like I'm blocked by this, I'm just trying to help out as much as I can to help fix the issues caused when I had to change the import path :).

michelle192837:
Makes sense. ^^ Thanks!

PR #238, opened 2020-09-29
Add DashboardTabFlakinessAlertOptions to config.proto

RonWeber:
Add a configuration setting for the flakiness alerter (Testgrid will eventually alert about flaky tests).


michelle192837 (approved):

pb/config/config.proto
```diff
@@ -543,6 +543,9 @@ message DashboardTab {
   // Configuration options for dashboard tab alerts.
   DashboardTabAlertOptions alert_options = 15;
 
+  // Configuration options for dashboard tab flakiness alerts.
+  DashboardTabFlakinessAlertOptions flakiness_alert_options = 24;
```
e-blackwelder, line 547: Is there a reason to not put this in numerical order? (I would expect future maintainers to simply scroll down to the end of the message and say "looks like 23 is the last one".)

RonWeber: It seemed unintuitive to separate alert_options and flakiness_alert_options.

michelle192837 (approved):

PR #239, opened 2020-09-30
Alerter proto update

chases2:
Adds alerter information to summary, to allow alerters to store timestamps

Also includes another field to be migrated in alerter consolidation


pb/summary/summary.proto
```diff
@@ -122,6 +122,12 @@ message HealthinessInfo {
   repeated float previous_flakiness = 5;
 }
 
+// Information about alerts that have been sent
+message AlertingData {
+  // Seconds since epoch at which an email was last sent
+  double last_email_timestamp = 1;
```
michelle192837, line 128: This should be a timestamp instead of a double, if possible

michelle192837 (approved):
/hold

Unhold once comment is addressed

pb/summary/summary.proto
chases2: How difficult will it be to migrate from a double to a timestamp?

The equivalent entry we're trying to migrate this from, [`last_alert_mail_time`](https://github.com/GoogleCloudPlatform/testgrid/blob/4e72272c6d8e329a458c686056fc7d19fe12c786/pb/state/state.proto#L167), is a double for legacy reasons.

michelle192837: Should be relatively easy; the timestamp message just has 'seconds' and 'nanos' as fields, so instead of using the field directly, you'd use the timestamp.seconds instead (maybe multiplied by some factor because I don't remember if we store seconds, milliseconds, or something else ><)

chases2:
Suggested changes have been made
/hold cancel

pb/summary/summary.proto
chases2: Done

michelle192837 (approved):

Issue #236, opened 2020-09-22
redhat-openshift-ocp-release-4.6-informing has both timestamp and sequential IDs


fejta:
> https://testgrid.k8s.io/redhat-openshift-ocp-release-4.6-informing#release-openshift-ocp-installer-e2e-ovirt-4.6 has a different set of tests than https://testgrid.k8s.io/redhat-openshift-ocp-release-4.6-informing#release-openshift-ocp-installer-e2e-ovirt-4.6&grid=old even though they are the same job. Is this inconsistency expected?

ref https://github.com/GoogleCloudPlatform/testgrid/issues/211

```
time="2020-09-22T17:17:57Z" level=error msg="Error updating group" func=github.com/GoogleCloudPlatform/testgrid/pkg/updater.Update.func1 file="pkg/updater/updater.go:70" config="gs://k8s-testgrid/config" error="read columns: read gs://gs://origin-ci-test/logs/release-openshift-ocp-installer-e2e-ovirt-4.6/1308157780305121280/: timeout: context deadline exceeded" group=release-openshift-ocp-installer-e2e-ovirt-4.6
```

https://github.com/GoogleCloudPlatform/testgrid/blob/18c11b364d36cf7cec42f8ca8bc1265eabfce942/cmd/updater/main.go#L82

/assign

fejta:
Think I finally have this fixed: https://external-canary-dot-k8s-testgrid.appspot.com/redhat-openshift-ocp-release-4.6-informing#release-openshift-ocp-installer-e2e-ovirt-4.6&grid=beta

fejta:
No, it actually has to do with https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/release-openshift-ocp-installer-e2e-ovirt-4.6/ having some really small numbers

PR #255, opened 2020-10-19
Handle GCS prefix returning earlier builds

fejta:
Also:
* Do not duplicate `gs://` when printing paths (this is already part of the build path's URL scheme)
* Add unit test coverage for this scenario
* Make unit tests more robust (the linking fake was cheating a bit too much)

ref https://github.com/GoogleCloudPlatform/testgrid/issues/236

### DIFF
--- a/util/gcs/read.go
+++ b/util/gcs/read.go
@@ -58,7 +58,7 @@ type Build struct {
 }
 
 func (build Build) String() string {
-	return "gs://" + build.Path.String()
+	return build.Path.String()
 }
 
 // ListBuilds returns the array of builds under path, sorted in monotonically decreasing order.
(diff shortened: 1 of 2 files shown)


util/gcs/read.go
```diff
@@ -58,7 +58,7 @@ type Build struct {
 }
 
 func (build Build) String() string {
-	return "gs://" + build.Path.String()
+	return build.Path.String()
```
chases2, line 61: I wonder how long it's been part "gs://" has been part of the build path's scheme. Has this function just been returning "gs://gs://path" and some later bit of code has been tolerating it?

chases2 (approved):
Looks good to me! Merge when ready.
/hold

util/gcs/read.go
fejta: I believe it is only used for debugging. So something is probably logging gs://gs://foo but this is never actually getting otherwise used.

I suspect that when I initially wrote this the gcs.Path didn't include the gs:// part on its debug String() and then later added it.

PR #257, opened 2020-10-20
Improve updater documentation

fejta:


cmd/updater/README.md
```diff
@@ -1,7 +1,63 @@
 
-The testgrid updater reads results from GCS to create a state proto.
+# Testgrid Updater
 
-The testgrid server reads these protos, converts them to json which the
+This component is responsible for compiling collated GCS results into a state proto.
```
e-blackwelder, line 4: Nit: hyperlinking to the [state.proto](https://github.com/GoogleCloudPlatform/testgrid/blob/master/pb/state/state.proto) file would be cool.

e-blackwelder (approved):

PR #263, opened 2020-10-30
Fix function comment and capitalization.

michelle192837:


chases2 (approved):
/hold
I think a comment, perhaps by the statusSeverity map, that gives meaning to "less" and "greater" is important.

Right now it looks like it's just an enum, but I think it's a severity or a precedence ordering.

fejta (approved):

fejta:
/hold cancel
Want to cut a release with this method signature change. Michelle can you address the doc change with a follow up?

PR #264, opened 2020-10-31
Refactor update to allow control over updateGroup.

fejta:
Intent is to allow private updaters more control over the update process.


e-blackwelder (approved):
Think I get the idea. LGTM

PR #266, opened 2020-11-02
Fix IsPassingResult() and Coalesce().

fejta:
BUILD_PASSED should be a passing result.
Coalesce should consider anything not passing (and not a special case) a failure.

/assign @michelle192837

### DIFF
--- a/internal/result/results.go
+++ b/internal/result/results.go
@@ -51,26 +51,26 @@ var (
 	}
 )
 
-// LTE returns if rowResult is less than or equal to compareTo.
-func LTE(rowResult, compareTo statuspb.TestStatus) bool {
+// lte returns if rowResult is less than or equal to compareTo.
+func lte(rowResult, compareTo statuspb.TestStatus) bool {
 	return statusSeverity[rowResult] <= statusSeverity[compareTo]
 }
 
-// GTE returns if rowResult is greater than or equal to compareTo.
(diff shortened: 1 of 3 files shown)


fejta:
Okay, finally got this one passing

internal/result/results.go
```diff
@@ -51,26 +51,26 @@ var (
...
 
-// LTE returns if rowResult is less than or equal to compareTo.
-func LTE(rowResult, compareTo statuspb.TestStatus) bool {
+// lte returns if rowResult is less than or equal to compareTo.
+func lte(rowResult, compareTo statuspb.TestStatus) bool {
 	return statusSeverity[rowResult] <= statusSeverity[compareTo]
 }
 
-// GTE returns if rowResult is greater than or equal to compareTo.
-func GTE(rowResult, compareTo statuspb.TestStatus) bool {
```
michelle192837, line 60: Can we keep this exported for internal use, or should we just have it exported there?

michelle192837 (approved):
/hold

Unhold after comment addressed!

internal/result/results.go
fejta: It wasn't exported historically? Does something need to use it internally?

fejta:
/hold cancel

Will send a subsequent PR if we need this exported

PR #270, opened 2020-11-06
Allow local defaults for a config directory.

michelle192837:
Test: Added unit tests, locally tested that default.yaml applies
correctly in k8s/test-infra config.

Note: We will need to later update the version of TestGrid used
in k8s/test-infra to pull these changes in.

See #7 

Co-authored-by: Mitchel <mpherman@google.com>

### DIFF
--- a/config/yamlcfg/yaml2proto.go
+++ b/config/yamlcfg/yaml2proto.go
@@ -28,15 +28,71 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
+// getDefaults take all paths found through seeking, returns list of dirs with defaults
+func getDefaults(allPaths []string) (defaults []string, err error) {
+	dirs_found := make(map[string]bool)
+	for _, path := range allPaths {
+		if filepath.Base(path) == "default.yaml" || filepath.Base(path) == "default.yml" {
+			if _, ok := dirs_found[filepath.Dir(path)]; ok {
+				return nil, fmt.Errorf("two default files found in dir %q", filepath.Dir(path))
+			}
+			defaults = append(defaults, path)
+			dirs_found[filepath.Dir(path)] = true
+		}
+	}
+	return defaults, nil
+}
+
+// seekDefaults finds all default files and returns a map of directory to its default contents.
+// TODO: Implement filesystem fake in order to unit test this better.
+func seekDefaults(paths []string) (map[string]DefaultConfiguration, error) {
+	defaultFiles := make(map[string]DefaultConfiguration)
+	var allPaths []string
+	err := SeekYAMLFiles(paths, func(path string, info os.FileInfo) error {
+		allPaths = append(allPaths, path)
+		return nil
+	})
+	if err != nil {
+		return nil, fmt.Errorf("unable to walk paths, %v", err)
+	}
+	defaults, err := getDefaults(allPaths)
+	if err != nil {
+		return nil, fmt.Errorf("unable to get defaults, %v", err)
+	}
+	for _, path := range defaults {
+		b, err := ioutil.ReadFile(path)
+		if err != nil {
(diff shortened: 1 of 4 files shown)


michelle192837:
@googlebot I fixed it.

config/yamlcfg/yaml2proto_test.go
```diff
@@ -169,6 +170,7 @@ test_groups:
 			expectedDashTab:   20,
 		},
 		{
+			// TODO: Do we want to prevent this case?
```
chases2, line 173: Prevent as in "fail if there are defaults specified outside of default.yaml"?

I think that might be clearer than warning that they do nothing (current behavior). If this is done with a clear enough error ("Please delete defaults in *.yaml, they don't do anything"), I'd support it.

Consider changing this TODO to "Yes, let's prevent this actually" and implementing in a separate PR.

config/yamlcfg/yaml2proto_test.go
```diff
@@ -710,3 +712,103 @@ func Test_ReadConfig(t *testing.T) {
...
+			name:  "simple case",
+			paths: []string{"foo/config.yaml", "foo/default.yaml"},
+			want:  []string{"foo/default.yaml"},
+		},
+		{
+			name:  "no defaults",
+			paths: []string{"foo/config.yaml"},
+		},
+		{
+			name:  "multiple defaults",
```
chases2, line 735: What happens if `default.yaml` and `default.yml` exist together?

config/yamlcfg/yaml2proto_test.go
```diff
@@ -710,3 +712,103 @@ func Test_ReadConfig(t *testing.T) {
...
+			name: "path not in map uses overall defaults",
+			path: "config.yaml",
+			defaultFiles: map[string]DefaultConfiguration{
+				"foo": localDefaults,
+			},
+			defaults: overallDefaults,
+			want:     overallDefaults,
+		},
+		{
+			name: "path in subdirectory of path in map uses overall defaults",
```
chases2, line 797: nit: "path in subdirectory of local default uses overall defaults"

It's important to me that this description be clear, since I might expect the default closest to the config file, not the root default.

chases2 (commented):
Looks great! Thank you!
/approve

config/yamlcfg/yaml2proto_test.go
mpherman2: Done

config/yamlcfg/yaml2proto_test.go
mpherman2: Done

config/yamlcfg/yaml2proto_test.go
mpherman2: Done

config/yamlcfg/yaml2proto.go
```diff
@@ -28,15 +28,71 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
+// getDefaults take all paths found through seeking, returns list of dirs with defaults
+func getDefaults(allPaths []string) (defaults []string, err error) {
+	dirs_found := make(map[string]bool)
+	for _, path := range allPaths {
+		if _, ok := map[string]bool{"default.yaml": true, "default.yml": true}[filepath.Base(path)]; ok {
```
chases2, line 35: `if filepath.Base(path) == "default.yaml" or filepath.Base(path) == "default.yml"`, right?

I'm having trouble parsing this map literal or why it's there.

michelle192837: Ah, map literal was us discovering a map can't be a constant, and deciding not to use a list instead. I agree the explicit OR makes more sense, done!

chases2 (approved):

PR #271, opened 2020-11-06
Update alerts in protos with hotlist IDs.

michelle192837:
This will let us write these properties in the updater.


chases2:
Why and/or how are we collecting this data with the updater? I'm missing some context here.

michelle192837:
Np! I'm doing this to enable keeping relevant values from `hotlist_ids_from_source`; need to gather it in the Updater and get it to the Summarizer in order to have it available for the new AutoBug component.

chases2 (approved):

PR #273, opened 2020-11-17
Change config to reject invalid fields for yaml

mpherman2:
...to prevent people from uploading configs that have incorrect fields and don't do anything. For example: default_test_group in a non default file.

### DIFF
--- a/config/yamlcfg/yaml2proto.go
+++ b/config/yamlcfg/yaml2proto.go
@@ -88,7 +88,7 @@ func pathDefault(path string, defaultFiles map[string]DefaultConfiguration, defa
 // 		 after applying defaults from defaultPath.
 // Optionally, defaultPath points to default setting YAML
 // Returns a configuration proto containing the data from all of those sources
-func ReadConfig(paths []string, defaultpath string) (config.Configuration, error) {
+func ReadConfig(paths []string, defaultpath string, strict bool) (config.Configuration, error) {
 
 	var result config.Configuration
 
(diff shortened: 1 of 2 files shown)


mpherman2:
friendly ping on re-review

fejta:
> friendly ping on re-review

Sorry, have been waiting for you to address the comment one way or another

fejta (approved):
So I'm vaguely concerned that this violates the robustness principle: https://en.wikipedia.org/wiki/Robustness_principle

> _Be conservative in what you send, be liberal in what you accept_



Specifically I would prefer that:
* On presubmit we ensure that are no unknown fields
* In production we ignore unknown fields

/hold

fejta:
Also temporarily placing both `old_field_name` and `new_field_name` in a doc is a common way we handle migrations that will work on both the old and new version

fejta:
(BTW my intent with lgtm + `/hold` is just for you to control when it merges with an `/hold cancel` command)

mpherman2:
> (BTW my intent with lgtm + `/hold` is just for you to control when it merges with an `/hold cancel` command).
> 
> AKA it allows me to be more liberal with my feedback while allowing you to control whether you address them in this PR, the next one (or never)

I appreciate the clarification :) I wasn't sure what the protocol was for lgtm + hold. Michelle and I are going to try and find some time to pair program and address these before we submit

mpherman2:
> So I'm vaguely concerned that this violates the robustness principle: https://en.wikipedia.org/wiki/Robustness_principle
> 
> > _Be conservative in what you send, be liberal in what you accept_
> 
> Specifically I would prefer that:
> 
> * On presubmit we ensure that are no unknown fields
> * In production we ignore unknown fields
> 
> /hold

We changed Update so that it only optionally uses StrictUnmarshall This will require some changes to k8s/t-i to use the new parameters for ReadConfig when we update testgrid in k8s/t-i

michelle192837 (approved):

fejta (approved):

Issue #211, opened 2020-08-17
Preview the golang updater results


fejta:
You can now see the results of the open-source golang updater processing your tabs (it will default to the python one).

Please link here to any odd tabs (and what is odd about them).

Toggle between the prod (aka python) and beta (aka open source golang) versions of the data by appending `&grid=beta` to the URL fragment.

AKA navigate somewhere at https://testgrid.k8s.io, open up a tab and then add `&grid=beta` to the end of the address bar, then by back by deleing this parameter and/or changing it to `&grid=old`:
* [Example](https://testgrid.k8s.io/sig-testing-misc#local-up-cluster,%20master%20(dev)&grid=beta)

fejta:
Sent out notice that this is available: https://groups.google.com/g/kubernetes-dev/c/dF5h3XuDV7k

detiber:
I tried viewing the Cluster API e2e tests using the new updater and it looks like it isn't showing the junit results broken out:
https://testgrid.k8s.io/sig-cluster-lifecycle-cluster-api#capi%20e2e%20tests&grid=old vs https://testgrid.k8s.io/sig-cluster-lifecycle-cluster-api#capi%20e2e%20tests&grid=beta

fejta:
Thanks for the report @detiber ! Corrected on canary (rolling out to prod soon): https://external-canary-dot-k8s-testgrid.appspot.com/sig-cluster-lifecycle-cluster-api#capi%20e2e%20tests&grid=beta

tehcyx:
Hey @fejta,
Comparing https://testgrid.k8s.io/kyma_integration#post-kyma-release-upgrade with https://testgrid.k8s.io/kyma_integration#post-kyma-release-upgrade&grid=beta

I'm missing all the details of the individual tests of the suite that was run. Assuming it's the same or similar to what @detiber posted above.

fejta:
This fix for this landed in production this afternoo, looks like things look better? https://testgrid.k8s.io/kyma_integration#post-kyma-release-upgrade&grid=beta

tehcyx:
> This fix for this landed in production this afternoo, looks like things look better? https://testgrid.k8s.io/kyma_integration#post-kyma-release-upgrade&grid=beta

Indeed, in the old one there's still a deep link to the changes, as well as this note `Show 2 stale tests (no results in last 10+ runs)` with a link that doesn't seem to change whatever is displayed?

fejta:
[...]

PR #277, opened 2020-12-03
Find link in either link or x-goog-meta-link header

fejta:
ref https://github.com/kubernetes/test-infra/issues/19696 https://github.com/GoogleCloudPlatform/testgrid/issues/211


michelle192837 (approved):
Awesome, thank you Erick! :D

PR #279, opened 2020-12-04
Run canary updater less frequently.

fejta:
Still runs immediately after each update

/hold


chases2 (approved):

fejta:
ref https://github.com/kubernetes/test-infra/issues/20010

PR #308, opened 2020-12-17
Fix some lint errors

fejta:


util/gcs/client.go
```diff
@@ -55,7 +55,7 @@ type Stater interface {
 	Stat(ctx context.Context, prefix Path) (*storage.ObjectAttrs, error)
 }
 
-// A copier can cloud copy an object to a new location if it
+// A Copier can cloud copy an object to a new location if it
```
chases2, line 58: if it... what?

chases2 (approved):
Looks good to me; one incomplete comment that might be worth fixing
/hold

util/gcs/client.go
fejta: Done

Issue #312, opened 2020-12-22
ppc64le-unit has large numbers of empty rows


fejta:
https://k8s-testgrid.appspot.com/sig-node-ppc64le#unit-tests why does this show 27k stale rows, which seem to have never had any results?

/assign

fejta:
ref https://github.com/kubernetes/test-infra/issues/20010

Rajalakshmi-Girish:
@fejta I now see the results updated till 21-12-2020, unlike yesterday only till 11-12-2020. Please let me know if anything changed that caused these runs to be updated?

Also, the 27k stale rows are due to the "periodic-kubernetes-bazel-test-ppc64le" prow jobs that are stuck in running state for a long time. I shall look into this.

@mkumatag

fejta:
Yes https://github.com/GoogleCloudPlatform/testgrid/pull/311 which had a mitigation for this scenario (a very wide and tall grid)

fejta:
> "periodic-kubernetes-bazel-test-ppc64le" prow jobs that are stuck in running state for a long time

+1 to looking your looking into any issues you may be having with your jobs but I do think this is a testgrid processing issue (note that they do not appear on &grid=old, right?)

fejta:
Yep, nothing wrong on your end (we already handle the scenario where jobs never finish, we just need to drop rows that get renamed, see linked PR)

Rajalakshmi-Girish:
> 
> 
> > "periodic-kubernetes-bazel-test-ppc64le" prow jobs that are stuck in running state for a long time
> 
> +1 to looking your looking into any issues you may be having with your jobs but I do think this is a testgrid processing issue (note that they do not appear on &grid=old, right?)

The empty rows do appear even in &grid=old. Please see below:
![image](https://user-images.githubusercontent.com/63038004/102886104-81d74600-447a-11eb-9cc3-04d543f32e0c.png)

fejta:
That shows a column that never finished, with a most of the rows empty on that particular column.

The problem here was the "Show 2 stale tests" link, which was actually showing 27000 rather than 2 (see [here](https://github.com/kubernetes/test-infra/issues/20010#issuecomment-748937585)). 

The 2 stale results are also not a problem, if you expand and scroll back to Dec 11 you'll see that //test/e2e/storage/testsuites possibly moved to //test/e2e/storage/utils -- eventually those two will clean up.

[...]

PR #313, opened 2020-12-22
Remove rows that no longer appear on the grid

fejta:
ref https://github.com/GoogleCloudPlatform/testgrid/issues/312


chases2:
```
ERROR: found unformatted go files, fix with:
  bazel run @io_k8s_repo_infra//hack:update-gofmt
```

chases2 (approved):

PR #327, opened 2021-01-15
Allow groups to collate multiple prow jobs together.

fejta:
Restrict this to specific groups as this is not a fully-finished feature

Switch `originalPrefix` to `baseName` is this is the reason we track it.
Export `gcs.Sort()` so that we can re-sort builds from multiple jobs

### DIFF
--- a/pkg/updater/updater.go
+++ b/pkg/updater/updater.go
@@ -29,6 +29,7 @@ import (
 	"path"
 	"runtime"
 	"sort"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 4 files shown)


pkg/updater/updater.go
```diff
@@ -274,20 +275,41 @@ func logUpdate(ch <-chan int, total int, msg string) {
...
 
-func groupPath(tg configpb.TestGroup) (*gcs.Path, error) {
-	u, err := url.Parse("gs://" + tg.GcsPrefix)
-	if err != nil {
-		return nil, fmt.Errorf("parse: %w", err)
-	}
-	if u.Path != "" && u.Path[len(u.Path)-1] != '/' {
-		u.Path += "/"
-	}
+// AllowMultiplePaths enables combining multiple jobs together using a comman.
```
michelle192837, line 278: nit:comma?

michelle192837 (approved):
/hold

pkg/updater/updater.go
fejta: Done
/hold cancel

michelle192837 (approved):

PR #344, opened 2021-01-27
Add job name when results collate multiple jobs

fejta:
If `GcsPrefix: this-job,that-job` then ensure that the name config includes a `Job name` field (insert it to the front if not)

### DIFF
--- a/pkg/updater/read.go
+++ b/pkg/updater/read.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"path"
+	"strings"
 	"sync"
 	"time"
 
@@ -112,7 +113,7 @@ func readColumns(parent context.Context, client gcs.Downloader, group *configpb.
 	// Concurrently receive indices and read builds
 	wg.Add(concurrency)
 	for i := 0; i < concurrency; i++ {
-		nameCfg := makeNameConfig(group.TestNameConfig)
+		nameCfg := makeNameConfig(group)
 		go func() {
 			defer wg.Done()
 			for {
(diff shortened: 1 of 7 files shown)


pkg/updater/read.go
```diff
@@ -208,6 +209,50 @@ func readColumns(parent context.Context, client gcs.Downloader, group *configpb.
 	return cols[0:maxIdx], nil
 }
 
+type nameConfig struct {
```
chases2, line 212: nameConfig seems like an important part of understanding this code, but I don't quite get what it is. It's a... "decompiled" representation of how test groups are named based on their config settings? Are the parts the properties to look for, or the properties themselves?

Some comments on this struct would help a lot.

util/gcs/read_test.go
```diff
@@ -49,6 +49,42 @@ func link(path Path, name, other string) storage.ObjectAttrs {
...
 }
 
+func TestBuildJob(t *testing.T) {
+	cases := []struct {
+		path  string
+		build string
+		job   string
+	}{
+		{
+			path:  "gs://bucket/path/job/hello",
```
chases2, line 59: I'm starting to see a pattern here of files used by controller `foo` are generally put into a subfolder called `foo`. I like that organization.

chases2 (approved):
/hold
/lint

This does seem clearer, but I'm still double-guessing while reading this code. A few comments on unexported functions might help a lot.

pkg/updater/read.go
fejta: This struct is the same as before, now in a more appropriate location.

I generally think the best way to understand code -- especially package internal code -- is to read the code.

PR #349, opened 2021-01-28
Create config merger controller

chases2:
Creates a controller that merges configs from multiple sources.

Written with many users in mind across multiple repositories and teams, such as teams currently using [Transfigure](https://github.com/kubernetes/test-infra/tree/master/testgrid/cmd/transfigure)

/assign @e-blackwelder 
/lint


cmd/config_merger/main.go
```diff
@@ -0,0 +1,187 @@
...
+	for range timer.C {
+		timer.Reset(opt.wait)
+		if err := updateOnce(ctx); err != nil {
+			logrus.WithError(err).Error("Failed update")
+		}
+		logrus.WithField("wait", opt.wait).Info("Sleeping")
+	}
+}
+
+func mergeAndUpdate(ctx context.Context, client *storage.Client, paths map[string]*gcs.Path, targetPath gcs.Path, confirm bool) {
```
fejta, line 152: Will you just use a gcs.Client? It includes an uploader:

https://github.com/GoogleCloudPlatform/testgrid/blob/0180bbe30254944d8f3cdd1861e242d81afc6d76/util/gcs/client.go#L65-L66

Would also allow unit testing, which I think is a good idea.

fejta: https://github.com/GoogleCloudPlatform/testgrid/blob/0180bbe30254944d8f3cdd1861e242d81afc6d76/pkg/updater/updater_test.go#L91-L245 an example that includes faking an upload as well as a fake config download

chases2: Oh, this helps to explain the distinction between the `cmd/` and `pkg/` directories

I've done part of the refactor, and will move the mergeAndUpdate() function and write tests for it next.

chases2: Refactor completed. The configuration and merge functions are in the `pkg/` directory with tests demonstrating when an upload is and isn't performed.

pkg/merger/merger.go
```diff
@@ -0,0 +1,129 @@
...
+	// TODO: Cache the version for each source. Only read if they've changed.
+	shards := make(map[string]*configpb.Configuration)
+
+	for key, path := range paths {
+		if path == nil {
+			return fmt.Errorf("path at '%s' is nil", key)
+		}
+		cfg, err := config.ReadGCS(ctx, client, *path)
+		if err != nil {
+			return fmt.Errorf("Can't read config %s at %v: %w.", key, path, err)
```
michelle192837, line 95: nit: Think these and other errors shouldn't be capitalized/punctuated

michelle192837 (approved):
/hold

pkg/merger/merger.go
chases2: Fixed

cmd/config_merger/main.go
```diff
@@ -0,0 +1,109 @@
...
+
+func main() {
+	opt := gatherOptions()
+	err := opt.validate()
+	if err != nil {
+		logrus.WithError(err).Fatal("Invalid flags")
+	}
+	file, err := ioutil.ReadFile(opt.listPath)
+	if err != nil {
+		logrus.WithError(err).Fatalf("Can't read file at %s", opt.listPath)
```
fejta, line 71: logrus.WithField("--config-list", opt.ListPath).WithError(err).Fata("Cannot find config")

cmd/config_merger/main.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+	if o.skipValidate {
+		logrus.Info("--allow-invalid-configs: result may not validate either")
+	}
+	return nil
+}
+
+func gatherOptions() options {
+	var o options
+	flag.StringVar(&o.listPath, "config-list", "", "List of configurations to merge")
```
fejta, line 54: Kind of surprised this refers to a typed yaml document, rather than a comma delimited list of config paths and/or a repeatable flag.

pkg/merger/merger_test.go
```diff
@@ -0,0 +1,324 @@
+/*
+Copyright 2020 The Kubernetes Authors.
```
fejta, line 2: 2021 for all these new files, right?

pkg/merger/merger.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	buf, err := proto.Marshal(result)
+	if err != nil {
+		return fmt.Errorf("failed to serialize merged proto: %w", err)
+	}
+
+	if confirm {
+		err = client.Upload(ctx, targetPath, buf, gcs.DefaultAcl, "no-cache")
+		if err != nil {
```
fejta, line 123: combine with previous line:

```golang
// since this is inside a {} you can also do if err := client.Upload()
if err = client.Upload(...); err != nil {
  return fmt.Errorf("could not...")
}
```

pkg/merger/merger.go
```diff
@@ -0,0 +1,129 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to merge configurations: %w", err)
+	}
+
+	buf, err := proto.Marshal(result)
+	if err != nil {
+		return fmt.Errorf("failed to serialize merged proto: %w", err)
+	}
+
+	if confirm {
```
fejta, line 121: How about printing it out if `!confirm` (`fmt.Println(result)` should look reasonable)?

pkg/merger/merger.go
```diff
@@ -0,0 +1,129 @@
...
+	if err != nil {
+		return
+	}
+
+	target, err = gcs.NewPath(cfg.Target)
+	if err != nil {
+		return
+	}
+
+	sources = make(map[string]*gcs.Path)
```
fejta, line 56: This seems odd. Were you getting panics in your unit tests or something? If so I would fix by something like:

```golang
if tc.sources == nil {
  tc.sources = map[string]*gcs.Path{}
}
```

pkg/merger/merger.go
```diff
@@ -0,0 +1,129 @@
...
+			return nil, nil, fmt.Errorf("duplicated name %s", source.Name)
+		}
+		path, err := gcs.NewPath(source.Location)
+		if err != nil {
+			return nil, nil, err
+		}
+		sources[source.Name] = path
+	}
+
+	if len(sources) == 0 {
```
fejta, line 68: Seems like we can do this earlier by checking `len(cfg.Sources) == 0`, right?

fejta (commented):
This is super cool!

A few comments but they are almost entirely minor cosmetic/style type things.

cmd/config_merger/main.go
chases2: I really wanted to include at least a "contact" field for entries, since a target use case for this is multiple teams merging their configurations onto one TG instance.

If one team's configuration is deleted or corrupted, we need a way to surface that to them as soon as possible without breaking every other tenant. Ideally, their presubmits would prevent this, but "please give us a way to alert you" seems like a good precaution. 

I also expect ParseAndCheck to be sufficient as a presubmit to check this yaml config file for obvious errors, but it's not obvious that the data at a valid path is invalid until you actually check.

pkg/merger/merger.go
chases2: Added this to the revision

pkg/merger/merger.go
chases2: Now that this is a set used to speed up name duplication detection, this should be less odd.

pkg/merger/merger.go
chases2: Done

pkg/merger/merger.go
chases2: Done

pkg/merger/merger_test.go
chases2: Done

pkg/merger/merger.go
```diff
@@ -0,0 +1,129 @@
...
+
+// MergeAndUpdate gathers configurations from each path and merges them.
+// Puts the result at targetPath if confirm is true
+// Will skip an input config if it is invalid and skipValidate is false
+// Other problems are considered fatal and will return an error
+func MergeAndUpdate(ctx context.Context, client mergeClient, paths map[string]*gcs.Path, targetPath gcs.Path, skipValidate, confirm bool) error {
+	// Deserialize each proto
+	// TODO: Reading and validating can be done in parallel with a wait group.
+	// TODO: Cache the version for each source. Only read if they've changed.
+	shards := make(map[string]*configpb.Configuration)
```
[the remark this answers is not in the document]
chases2: Done throughout; I should have all the make() calls out of this code now

cmd/config_merger/main.go
chases2: Done here and with similar logging calls

pkg/merger/merger.go
```diff
@@ -0,0 +1,129 @@
...
+type mergeClient interface {
+	gcs.Opener
+	gcs.Uploader
+}
+
+// MergeAndUpdate gathers configurations from each path and merges them.
+// Puts the result at targetPath if confirm is true
+// Will skip an input config if it is invalid and skipValidate is false
+// Other problems are considered fatal and will return an error
+func MergeAndUpdate(ctx context.Context, client mergeClient, paths map[string]*gcs.Path, targetPath gcs.Path, skipValidate, confirm bool) error {
```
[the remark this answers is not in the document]
chases2: Done

fejta (approved):
/hold

PR #372, opened 2021-02-20
Add local filesystem capability to util/gcs.

michelle192837:
Have checked that local files work with Merger and Updater (though not Summarizer as-is; that will need an update to how the storage works, will tackle in a follow-up PR).


util/gcs/gcs.go
```diff
@@ -75,8 +79,8 @@ func (g *Path) SetURL(u *url.URL) error {
 	switch {
 	case u == nil:
 		return errors.New("nil url")
-	case u.Scheme != "gs":
-		return fmt.Errorf("must use a gs:// url: %s", u)
+	case u.Scheme != "gs" && u.Scheme != "":
```
chases2, line 82: nit: consider tolerating `u.Scheme == "file"`, ex. "file:///tmp/config"

chases2 (approved):
The polymorphism here looks good to me. May need to do a BUILD.bazel update?
/lint

util/gcs/client.go
```diff
@@ -76,60 +75,54 @@ type ConditionalClient interface {
 	If(read, write *storage.Conditions) ConditionalClient
 }
 
-// NewClient returns a GCSUploadClient for the storage.Client.
-func NewClient(client *storage.Client) ConditionalClient {
-	return realGCSClient{client, nil, nil}
+type GCSClient struct {
```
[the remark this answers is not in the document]
fejta: No need to export this structure.

fejta:
Please git add the new files!

util/gcs/gcs.go
michelle192837: Done I think!

chases2 (approved):

PR #376, opened 2021-02-24
Have Summarizer use util/gcs library, not storage itself.

michelle192837:
Verified locally (e.g. `bazel run //cmd/summarizer:summarizer -- --config=/tmp/testgrid-states/config --confirm=true --dashboard=TestGrid`)

### DIFF
--- a/pkg/summarizer/summary.go
+++ b/pkg/summarizer/summary.go
@@ -55,11 +55,11 @@ type groupFinder func(string) (*configpb.TestGroup, gridReader, error)
 // Will use concurrency go routines to update dashboards in parallel.
 // Setting dashboard will limit update to this dashboard.
 // Will write summary proto when confirm is set.
-func Update(ctx context.Context, client *storage.Client, configPath gcs.Path, concurrency int, dashboard, gridPathPrefix, summaryPathPrefix string, confirm bool) error {
+func Update(ctx context.Context, client gcs.Client, configPath gcs.Path, concurrency int, dashboard, gridPathPrefix, summaryPathPrefix string, confirm bool) error {
 	if concurrency < 1 {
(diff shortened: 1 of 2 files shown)


pkg/summarizer/summary.go
```diff
@@ -156,21 +156,25 @@ func summaryPath(name string) string {
...
 }
 
 // pathReader returns a reader for the specified path and last modified, generation metadata.
-func pathReader(ctx context.Context, client *storage.Client, path gcs.Path) (io.ReadCloser, time.Time, int64, error) {
-	r, err := client.Bucket(path.Bucket()).Object(path.Object()).NewReader(ctx)
+func pathReader(ctx context.Context, client gcs.Client, path gcs.Path) (io.ReadCloser, time.Time, int64, error) {
+	r, err := client.Open(ctx, path)
 	if err != nil {
-		return nil, time.Time{}, 0, fmt.Errorf("read %s: %w", path, err)
+		return nil, time.Time{}, 0, fmt.Errorf("client.Open(%s): %w", path, err)
```
fejta, line 171: FYI I'd remove path from the returned error as it is trivial to the caller (aka the caller passes in `pathReader(..., path)` and therefore knows _exactly_ what it is and can decide whether or not to log/wrap it.

OTOH when the code is
```golang
for _, path := range paths {
  r, err := Open(path)
  if err != nil {
    return fmt.Errorf("Open(%q): %w", path, err)
  }
}
```

I _would_ include it because which of the many paths caused the problem is useful info.

fejta (approved):
/hold

pkg/summarizer/summary.go
michelle192837: Good point, thanks! Done.

fejta (approved):
/hold cancel

PR #381, opened 2021-02-25
Hack listing to handle sequential build numbers

fejta:
/assign @chaodaiG

### DIFF
--- a/util/gcs/read.go
+++ b/util/gcs/read.go
@@ -25,6 +25,7 @@ import (
 	"path"
 	"regexp"
 	"sort"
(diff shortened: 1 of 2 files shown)


chaodaiG (approved):
/hold
In case you want to fix the inclusive lang warnings

fejta:
I guess that's what I get for copying the actual job name! Can you restamp? The name doesn't actually matter, happy to just drop the word.

chaodaiG (approved):

michelle192837 (approved):

PR #382, opened 2021-02-26
Inject a Pod row into results

fejta:
This works similarly to the Overall row, except this data comes from podinfo.json.
Attempts to flag various types of pod problems that might prevent tests from completing.

(We already use kubernetes and therefore k8s.io/api inside Google, so the new import is not a problem)


e-blackwelder (approved):
Looks reasonable, but I'm not sure completely sure what's going on with multiJob and related logic (in gcs.go). Note: I didn't inspect every test case for appropriate use of e.g. podInfoPassCell vs podInfoMissingCell, etc.

/hold

michelle192837 (approved):

fejta:
Thanks!

> I didn't inspect every test case for appropriate use

For the most part the contents don't matter, but something does need to be there in those unit tests since this row is always added now.

> not sure completely sure what's going on with multiJob and related logic

For the moment we always want the Overall and Pod row to be called Pod and Overall, even when test methods are trying to customize how the name is formatted. OTOH in a multi-job situation we want this to become `job-foo.Pod` but not whatever their custom test name formatting.

/hold cancel

PR #395, opened 2021-03-04
Add quick script to compare two TestGrid states.

michelle192837:
This is not intended to compare two states in detail, but is a quick check that two states are pretty reasonably close by comparing rows and columns of each. (Also provides a diff of the rows and columns as it works).

This goes into hack/ as a possible helper script if you're making changes to Updater or something similar, and want to check states are not drastically wrong.

### DIFF
--- a/hack/compare_states.go
+++ b/hack/compare_states.go
@@ -1,0 +1,185 @@
+/*
+Copyright 2021 The TestGrid Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// A script to quickly check two TestGrid state.protos do not wildly differ.
+// Assume that if the column and row names are approx. equivalent, the state
+// is probably reasonable.
+
+package main
+
+import (
+	"context"
+	"errors"
+	"flag"
+	"fmt"
+	"github.com/GoogleCloudPlatform/testgrid/util/gcs"
+	"github.com/golang/protobuf/proto"
+	"github.com/google/go-cmp/cmp"
+	"github.com/sirupsen/logrus"
+	"google.golang.org/api/iterator"
+	"os"
+	"path/filepath"
+	"strings"
+
+	statepb "github.com/GoogleCloudPlatform/testgrid/pb/state"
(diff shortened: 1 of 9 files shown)


pkg/updater/read.go
```diff
@@ -36,7 +36,7 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-func downloadGrid(ctx context.Context, opener gcs.Opener, path gcs.Path) (*statepb.Grid, error) {
+func DownloadGrid(ctx context.Context, opener gcs.Opener, path gcs.Path) (*statepb.Grid, error) {
```
fejta, line 39: Perhaps we should just move this method into util/gcs?

hack/compare_states.go
```diff
@@ -0,0 +1,184 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// A script to quickly check two TestGrid state.protos do not wildly differ.
+// Assume that if the column and row names are approx. equivalent, the state
+// is probably reasonable.
+
+package main
```
fejta, line 21: Unit tests?

hack/compare_states.go
```diff
@@ -0,0 +1,184 @@
...
+		diffed, err := compare(ctx, *firstPath, *secondPath, client)
+		if err != nil {
+			errorMsgs = append(errorMsgs, fmt.Sprintf("compare(%q, %q): %v", firstP, secondP, err))
+			continue
+		}
+		if diffed {
+			diffedMsgs = append(diffedMsgs, fmt.Sprintf("%q vs. %q", firstP, secondP))
+		}
+	}
+	if len(errorMsgs) > 0 {
```
fejta, line 174: if n := len(errorMsgs); n > 0 { logrus.Error("Got %d errors:", n) }

hack/compare_states.go
```diff
@@ -0,0 +1,184 @@
...
+		logrus.Errorf("Got %d errors:", len(errorMsgs))
```
fejta, line 175: Consider something more like `logrus.WithField("count", len(errorMsgs)).Error("Directories are not the same")`

Also consider changing this to `logrus.Fatal()` and moving at the end so it exits non-zero on failure.

hack/compare_states.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+	return diffed, nil
+}
+
+func filenames(ctx context.Context, dir gcs.Path, client gcs.Client) []string {
+	stats := client.Objects(ctx, dir, "/", "")
+	var filenames []string
+	for {
+		stat, err := stats.Next()
+		if err != nil {
```
fejta, line 127: Should we distinguish `iterator.Done` from other errors? https://pkg.go.dev/google.golang.org/api/iterator#pkg-variables

fejta (approved):

Issue #393, opened 2021-03-03
Allow users to disable the Pod row via config


fejta:
The new Pod row analyzes the `podinfo.json` file that prow creates which reports the pod status, looking for any pod issues that might prevent the test container from starting/completing -- aka image pull errors, eviction issues, etc.

Not all results come from prow, and not all results run in a container, so testgrid should allow this functionality to be turned off.

ref https://github.com/kubernetes/test-infra/issues/21126

fejta:
Adding `disable_prowjob_analysis: true` on the test group will remove this row from future columns.

PR #396, opened 2021-03-05
Add ability to disable_prowjob_analysis.

fejta:
The updater generally assumes use_kubernetes_client type jobs come from prow,
and prow should be including diagnostic prowjob.json and podinfo.json. However
there are some results from other systems that create finished.json but do
not necessarily run in a pod and/or have a prowjob, so this row is irrelevant.

Allow such rows to set disable_prowjob_analysis: true to disable this row.

Add unit tests for this scenario, also test the disable_merged_status feature.

ref https://github.com/GoogleCloudPlatform/testgrid/issues/393


pkg/updater/gcs_test.go
```diff
@@ -1068,7 +1149,12 @@ func TestConvertResult(t *testing.T) {
 	for _, tc := range cases {
 		t.Run(tc.name, func(t *testing.T) {
 			log := logrus.WithField("test name", tc.name)
-			actual, err := convertResult(log, tc.nameCfg, tc.id, tc.headers, tc.metricKey, tc.result, tc.merge)
+			if tc.opt == nil {
```
chases2, line 1152: nit: I think I'd rather see this as an explicit add to these test cases

chases2 (approved):
/hold
Merge when ready; nits only

pkg/updater/gcs_test.go
fejta: sure! Working on a follw-up.
/hold cancel

PR #404, opened 2021-03-17
Validate dashboard name length and dashboardTab length

chases2:
Fixes https://github.com/GoogleCloudPlatform/testgrid/issues/98

- Tests added to check for empty dashboards and return an error in that case
- Tests added to check for a short entity length: 0-2 length entities are not valid

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -34,7 +34,8 @@ import (
 	multierror "github.com/hashicorp/go-multierror"
 )
 
-// MissingFieldError is an error that includes the missing field.
+// MissingFieldError is an error that includes the missing root field.
+// Entities that contain no children should use a ConfigError, so they can point to the empty Entity
 type MissingFieldError struct {
 	Field string
 }
@@ -81,6 +82,9 @@ func normalize(s string) string {
 	s = strings.ToLower(s)
 	return s
 }
+
+const MIN_NAME_LENGTH = 3
+const MAX_NAME_LENGTH = 2048
 
 // validateUnique checks that a list has no duplicate normalized entries.
 func validateUnique(items []string, entity string) error {
@@ -211,8 +215,13 @@ func validateReferencesExist(c *configpb.Configuration) error {
 // TestGrid file prefix, post-normalization.
 func validateName(s string) error {
 	name := normalize(s)
-	if name == "" {
-		return errors.New("normalized name can't be empty")
+
+	if len(name) < MIN_NAME_LENGTH {
+		return fmt.Errorf("names must contain at least %d alphanumeric characters", MIN_NAME_LENGTH)
+	}
+
+	if len(name) > MAX_NAME_LENGTH {
+		return fmt.Errorf("names should not contain more than %d alphanumeric characters", MAX_NAME_LENGTH)
 	}
 
 	invalidPrefixes := []string{"dashboard", "alerter", "summary", "bugs"}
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -211,8 +212,9 @@ func validateReferencesExist(c *configpb.Configuration) error {
 // TestGrid file prefix, post-normalization.
 func validateName(s string) error {
 	name := normalize(s)
-	if name == "" {
-		return errors.New("normalized name can't be empty")
+
+	if len(name) < 3 {
+		return errors.New("names must contain at least 3 alphanumeric characters")
```
fejta, line 217: Shouldn't there also be a maximum length?

config/config.go
```diff
@@ -443,6 +445,13 @@ func Validate(c *configpb.Configuration) error {
 		return multierror.Append(mErr, MissingFieldError{"Dashboards"})
 	}
 
+	// Each Dashboard must contain at least 1 Tab to do anything
+	for _, dashboard := range c.GetDashboards() {
+		if len(dashboard.DashboardTab) == 0 {
```
fejta, line 450: `dashboard.DashboardTab != ""` is the (faster and) more idiomatic way of writing this.

fejta (approved):
/hold

config/config.go
chases2: Maybe. What should it be?

config/config.go
chases2: dashboard.DashboardTab is a []*DashboardTab, not a string

config/config.go
fejta: I'd just pick a longish number and ensure it doesn't exceed this (such as 512, 1024 characters). Doesn't particularly matter.

config/config.go
```diff
@@ -211,8 +212,13 @@ func validateReferencesExist(c *configpb.Configuration) error {
...
 func validateName(s string) error {
 	name := normalize(s)
-	if name == "" {
-		return errors.New("normalized name can't be empty")
+
+	if len(name) < 3 {
+		return errors.New("names must contain at least 3 alphanumeric characters")
+	}
+
+	if len(name) > 2048 {
```
fejta, line 220: Add unit test?

I would make these consts for a more robust unit test, aka:

```golang
const small = 3, big = 4

cases := []struct{
  name string
  n int
  pass bool
}{
  {
    name: "too small fails",
    n: small-1,
    err: true,
  },
  {
    name: "small pass",
    n: small,
    err: true,
   },
   // same for big and big+1
}
```

also

```golang
var sb strings.Builder
for i := 0; i<tc.n;i++ { 
  sb.WriteRune('a')
}
name := sb.String()
```

fejta: (AKA that way we can change big from 2048 to 1024 or whatever without updating the unit test)
(Also fine to just use the 2048 literal in the unit test)

fejta (approved):
/hold
would like to see missing coverage added in either this PR or next one

config/config.go
chases2: This feels like overkill, but it does give me an opportunity to make existing tests a bit more descriptive.

config/config_test.go
```diff
@@ -363,37 +364,70 @@ func TestValidateReferencesExist(t *testing.T) {
 	}
 }
 
-func TestUpdate_validateNames(t *testing.T) {
+func TestValidateName(t *testing.T) {
+	stringOfLength := func(length int) string {
```
fejta, line 368: IMO this function doesn't need to exist as it is only called once right now.

In the future when someone is adding a new test to this list and/or debugging a failure that isn't the "Names can't be too long" test, this adds unnecessary complexity for that person, reducing their performance.

fejta (approved):
/hold

PR #406, opened 2021-03-23
Add JUnit "error" test cases

qinqon:
testgrid mark test as passed if they are of type "error" [1]. This
change add the type "error" and store it in the Result as Errored field.

[1] https://github.com/windyroad/JUnit-Schema/blob/master/JUnit.xsd#L75

Signed-off-by: Quique Llorente <ellorent@redhat.com>

### DIFF
--- a/metadata/junit/junit.go
+++ b/metadata/junit/junit.go
@@ -102,6 +102,7 @@ type Result struct {
 	Failure    *string     `xml:"failure,omitempty"`
 	Output     *string     `xml:"system-out,omitempty"`
 	Error      *string     `xml:"system-err,omitempty"`
+	Errored    *string     `xml:"error,omitempty"`
 	Skipped    *string     `xml:"skipped,omitempty"`
 	Properties *Properties `xml:"properties,omitempty"`
 }
@@ -130,10 +131,12 @@ func (r *Result) SetProperty(name, value string) {
 
 // Message extracts the message for the junit test case.
 //
-// Will use the first non-empty <failure/>, <skipped/>, <system-err/>, <system-out/> value.
+// Will use the first non-empty <error/>, <failure/>, <skipped/>, <system-err/>, <system-out/> value.
 func (r Result) Message(max int) string {
 	var msg string
 	switch {
+	case r.Errored != nil && *r.Errored != "":
+		msg = *r.Errored
 	case r.Failure != nil && *r.Failure != "":
 		msg = *r.Failure
 	case r.Skipped != nil && *r.Skipped != "":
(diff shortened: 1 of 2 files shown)


metadata/junit/junit.go
```diff
@@ -102,6 +102,7 @@ type Result struct {
 	Failure    *string     `xml:"failure,omitempty"`
 	Output     *string     `xml:"system-out,omitempty"`
 	Error      *string     `xml:"system-err,omitempty"`
+	Errored    *string     `xml:"error,omitempty"`
```
michelle192837, line 105: I think you want to process this in the Result.Truncate() function too.

(Also wondering how easy it is to rename 'Output' and 'Error' to append 'Sys' or 'System', but think that can be a separate PR.

qinqon: I will add the truncate, about the renaming, it will break compatibility, but If that's not a problem I can try to rename it.

michelle192837: I think leaving it to a separate PR (or filing an issue about it) is fine.

Also forgot, but you'll need to update Message() as well.

qinqon: Done

metadata/junit/junit.go
```diff
@@ -172,7 +175,7 @@ func truncatePointer(str *string, max int) {
 
 // Truncate ensures that strings do not exceed the specified length.
 func (r Result) Truncate(max int) {
-	for _, s := range []*string{r.Failure, r.Skipped, r.Error, r.Output} {
+	for _, s := range []*string{r.Error, r.Failure, r.Skipped, r.Output} {
```
michelle192837, line 178: Think this still needs to account for r.Errored too?

metadata/junit/junit.go
```diff
@@ -130,10 +131,12 @@ func (r *Result) SetProperty(name, value string) {
 
 // Message extracts the message for the junit test case.
 //
-// Will use the first non-empty <failure/>, <skipped/>, <system-err/>, <system-out/> value.
+// Will use the first non-empty <error/>, <failure/>, <skipped/>, <system-err/>, <system-out/> value.
 func (r Result) Message(max int) string {
 	var msg string
 	switch {
+	case r.Errored != nil && *r.Errored != "":
```
michelle192837, line 138: Add a corresponding test case to TestMessage().

metadata/junit/junit.go
qinqon: Made a mistake, fixed.

metadata/junit/junit.go
qinqon: Done

michelle192837 (approved):
Thanks! ^^

fejta:
These results will still show up as passing as testgrid until you update https://github.com/GoogleCloudPlatform/testgrid/blob/5714d1590e6c0968c1f4ab5d78a7ee6294625d9c/pkg/updater/gcs.go#L251-L255

PR #407, opened 2021-03-23
Add a test_method_unmatch_regex field to TestGroup

WillMorrison:


chases2 (approved):

chases2:
Hello! Thanks for your PR!

After modifying a proto, you'll want to run `bazel run //hack:update-protos` to regenerate the corresponding go code.

pb/config/config.proto
```diff
@@ -294,10 +292,15 @@ message TestGroup {
...
 
-  // Regex to match test methods. Only tests methods with names that match
+  // Regex to match test methods. Only test methods with names that match
   // this regex will be included in the table.
   string test_method_match_regex = 46;
 
+  // Regex to exclude test methods. Test methods with names that match
+  // this regex will be excluded from the table, even if they match
+  // test_method_match_regex.
+  string test_method_unmatch_regex = 59;
```
chases2, line 302: This number will need to be 61; it looks like 59 and 60 are reserved internally.

chases2: xref: https://github.com/GoogleCloudPlatform/testgrid/issues/26

WillMorrison: Done

PR #419, opened 2021-04-02
Only log the entire summary proto when confirm is false.

fejta:
Currently we are spamming our logs with the full content of every dashboard's summary every update loop.

### DIFF
--- a/pkg/summarizer/summary.go
+++ b/pkg/summarizer/summary.go
@@ -97,8 +97,8 @@ func Update(ctx context.Context, client gcs.Client, configPath gcs.Path, concurr
 					errCh <- errors.New(dash.Name)
 					continue
 				}
-				log.WithField("summary", sum).Info("summarized")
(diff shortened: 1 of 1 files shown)


pkg/summarizer/summary.go
```diff
@@ -97,8 +97,8 @@ func Update(ctx context.Context, client gcs.Client, configPath gcs.Path, concurr
 					errCh <- errors.New(dash.Name)
 					continue
 				}
-				log.WithField("summary", sum).Info("summarized")
 				if !confirm {
+					log.WithField("summary", sum).Info("Summarized")
```
michelle192837, line 101: Should we still log _that_ we summarized the dashboard in either case, and the full summary only when !confirm?

fejta: Done

chases2 (approved):
/hold

PR #429, opened 2021-04-10
Print Config Utility

chases2:
I wrote an experiment to see how well protos would deserialize into json objects (pretty well) and accidentally wrote a useful CLI config printing utility for debugging.

/lint

### DIFF
--- a/config/print/README.md
+++ b/config/print/README.md
@@ -1,0 +1,12 @@
+# Config Printer
+
+The config printer is a debugging utility that prints a TestGrid configuration in a
+human-readable format. It will read from the local filesystem or Google Cloud Storage.
+
+## Usage and installation
+
+```sh
+go install ./config/print
+print gs://example/config
+```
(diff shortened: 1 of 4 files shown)


config/print/README.md
```diff
@@ -0,0 +1,12 @@
...
+The config printer is a debugging utility that prints a TestGrid configuration in a
+human-readable format. It will read from the local filesystem or Google Cloud Storage.
+
+## Usage and installation
+
+```sh
+go install ./config/print
+print gs://example/config
+```
+A Bazel `run` command also works, but the utility won't be able to read from the local file system as expected.
```
fejta, line 12: https://github.com/kubernetes/test-infra/blob/47b6e0fed7cc39bf87b864302445fe381a9802dc/BUILD.bazel#L10

should address that

fejta (approved):
/hold

config/print/README.md
chases2: Neat! I'll mess with that and see if I can make a quick improvement later.

PR #430, opened 2021-04-21
Allow MergeCells to fail results with passes.

fejta:
Currently merging both passes and failures will always result in a flaky status.
Externally this is the correct behavior for junit prow results. However other
result systems need the ability to fail these types of results so add this ability.

Also take the highest filled message rather than the first non-empty one.

### DIFF
--- a/pkg/updater/gcs.go
+++ b/pkg/updater/gcs.go
@@ -101,7 +101,7 @@ const (
 	podInfoRow = "Pod"
 )
 
-func MergeCells(cells ...Cell) Cell {
+func MergeCells(flakyFail bool, cells ...Cell) Cell {
 	var out Cell
 	if len(cells) == 0 {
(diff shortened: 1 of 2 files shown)


pkg/updater/gcs.go
```diff
@@ -101,7 +101,7 @@ const (
 	podInfoRow = "Pod"
 )
 
-func MergeCells(cells ...Cell) Cell {
+func MergeCells(flakyFail bool, cells ...Cell) Cell {
```
michelle192837, line 104: Can you comment a bit on what flakyFail does?

michelle192837 (approved):

pkg/updater/gcs.go
fejta: #431 documents these functions (and chooses a better name for this argument)

PR #433, opened 2021-04-21
Modify TruncateRunning to truncate any running cell

chases2:
Modifies TruncateRunning to regenerate a row with any running job in it, not just the "Overall" row.

Allows for TestGrid to detect running jobs even when the "Overall" row isn't being injected.

### DIFF
--- a/pkg/updater/updater.go
+++ b/pkg/updater/updater.go
@@ -321,13 +321,15 @@ func truncateRunning(cols []InflatedColumn) []InflatedColumn {
 	if len(cols) == 0 {
 		return cols
 	}
-	var stillRunning int
-	for i, c := range cols {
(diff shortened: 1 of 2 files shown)


pkg/updater/updater.go
```diff
@@ -321,13 +321,15 @@ func truncateRunning(cols []InflatedColumn) []InflatedColumn {
 	if len(cols) == 0 {
 		return cols
 	}
-	var stillRunning int
-	for i, c := range cols {
-		if c.Cells[overallRow].Result == statuspb.TestStatus_RUNNING {
-			stillRunning = i + 1
+	for i := len(cols) - 1; i >= 0; i-- {
```
fejta, line 324: neat :)

fejta (approved):

fejta (approved):

Issue #441, opened 2021-04-27
sort summaries by last summarized time


fejta:
https://github.com/GoogleCloudPlatform/testgrid/blob/1c1a8a895cf4efa0299a39eb0a439b8b8ef1fe83/pkg/summarizer/summary.go#L138-L144

We should prioritize updating summaries by least recently summarized. That way if we fall behind we'll eventually get to all of them, rather than not updating the ones at the end of the list over and over

/assign

We do this in the updater and it works well: https://github.com/GoogleCloudPlatform/testgrid/blob/1c1a8a895cf4efa0299a39eb0a439b8b8ef1fe83/pkg/updater/updater.go#L72-L76

PR #443, opened 2021-04-28
 Summarize tab by least recently updated order. 

fejta:
Refactor `LeastRecentlyUpdated` into `util/gcs`
Move GCS fakes into `util/gcs/fake`
Refactor updater to use `LeastRecentlyUpdated`
Teach summarizer to use `LeastRecentlyUpdated`

Reviewing just the second commit should minimize eye strain (first commit moves the fakes)


fejta:
ref #441

michelle192837 (approved):

PR #445, opened 2021-04-28
Remove unused environment_instance field

fejta:


e-blackwelder (approved):
/lgtm
/hold
Looks Reasonable To Me: I don't remember seeing a "reserved" before, but I trust you're using it properly.

PR #446, opened 2021-04-29
Move Sort() into sort.go.

fejta:
Correctly document and add unit test

/assign @michelle192837

### DIFF
--- a/util/gcs/sort.go
+++ b/util/gcs/sort.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	"cloud.google.com/go/storage"
+	"github.com/fvbommel/sortorder"
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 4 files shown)


util/gcs/sort.go
```diff
@@ -93,3 +94,19 @@ func Touch(ctx context.Context, client ConditionalClient, path Path, generation
 	cond.DoesNotExist = true
 	return client.If(&cond, &cond).Upload(ctx, path, genZero, DefaultACL, "no-cache")
 }
+
+// Sort the builds by monotonically decreasing original prefix base name.
+//
+// In other words,
```
chases2, line 100: Please add this as a test case also. The current coverage doesn't make it clear how, for example, b/10 and c/1 should interact.

chases2 (approved):
/hold for author review

util/gcs/sort.go
fejta: Please read the other tests in this package, which already cover this scenario.

Ensuring that the `Build` for `gs://c/10` sets the `baseName` to 10 is not the responsibility of `Sort()`

fejta (commented):
/hold cancel

PR #448, opened 2021-04-30
Correctly group multiple columns together.

fejta:
* Fudge `stopTime` by an hour when a test group collates multiple jobs
  - Build Bx from job Jx may have an earlier start time but
    later build id than By from Jy.
  - If we already processed By (because of its smaller build id, then we
    currently reject processing Bx (since its start time is too small)
  - Fudge ensures we process both so long as they start within the same
    hour.
* Fix a truncating issue
  - Need to set to the end of the list since the order is descending.
* Fix `ListBuilds` to exclude a build matching offset.
* Drop `mergeColumns`
  - No longer necessary as `ListBuilds` works correctly
* Validate the sort is stable
  - By correctly implementing greater to also check that they are unequal

/assign @chizhg @michelle192837 @chases2


chizhg:
/lgtm
Thanks a lot @fejta !!!

Issue #461, opened 2021-05-04
No failures or alerts but still flaky


fejta:
https://testgrid.k8s.io/sig-windows-1.18-release

fejta:
It is flaky because:

* `num_columns_recent` is [configured](https://testgrid.k8s.io/q/config/sig-windows-1.18-release/aks-engine-windows-dockershim-serial-slow-1.18) to 10
* The `kubetest.Up` recently got renamed from `Up`
* The `Up` row is still present in the dashboard
  - The tenth most recent result for this column is a failure

So basically while the ten most recent rows are all passing, The tenth most recent cell for each row includes rows with failures, thus the summarizer considers this row flaky.

PR #462, opened 2021-05-04
Stop flaky/pass overall status after num_columns_recent filled columns.

fejta:
Currently we will consider num_columns_recent filled columns for each row. For sparse grids this
can result in an amusing side effect where all recent columns pass while the overall status
remains flaky. This is because a row may have failures in its num_columns_recent non-empty cells,
but if the empty cells are at the front, these failures may be outside the tab's num_columns_recent.

Fix this by stopping as soon as we find num_columns_recent columns with at least one filled cell.

Columns with running rows do not count toward num_columns_recent.

ref https://github.com/GoogleCloudPlatform/testgrid/issues/461

### DIFF
--- a/pkg/summarizer/summary.go
+++ b/pkg/summarizer/summary.go
@@ -584,11 +584,22 @@ func overallStatus(grid *statepb.Grid, recent int, stale string, brokenState boo
 	defer cancel()
 
 	results := results(ctx, grid.Rows)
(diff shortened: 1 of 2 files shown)


chases2 (approved):
A bit tricky to read, but I think moving the "break" up to the for loop does make this easier to understand.

PR #465, opened 2021-05-05
Add some documentation to overallStatus()

fejta:
/cc @chases2 

I am concerned with doing this sort of thing, because people tend to be more fluent in English than golang. Thus minds prefer reading the English. But only the golang determines how the program actually behaves. We need people maintaining this code to read the golang.

### DIFF
--- a/pkg/summarizer/summary.go
+++ b/pkg/summarizer/summary.go
@@ -571,6 +571,13 @@ func buildFailLink(testID, target string) string {
 }
 
 // overallStatus determines whether the tab is stale, failing, flaky or healthy.
+//
+// Tabs are:
+// BROKEN - called with brokenState (typically when most rows are red)
+// STALE - called with a stale mstring (typically when most recent column is old)
+// FAIL - there is at least one alert
+// FLAKY - at least one recent column has failing cells
+// PASS - all recent columns are entirely green
 func overallStatus(grid *statepb.Grid, recent int, stale string, brokenState bool, alerts []*summarypb.FailingTestSummary) summarypb.DashboardTabSummary_TabStatus {
 	if brokenState {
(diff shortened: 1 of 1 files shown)


fejta:
/retest 

flake https://github.com/GoogleCloudPlatform/testgrid/issues/466

pkg/summarizer/summary.go
```diff
@@ -571,6 +571,13 @@ func buildFailLink(testID, target string) string {
 }
 
 // overallStatus determines whether the tab is stale, failing, flaky or healthy.
+//
+// Tabs are:
```
chases2, line 575: In this example, this helps a whole lot. It tells me that you're trying to cover for 5 cases, and a bit about what each one means. Then, looking at the function, I see that 3 of the 5 are handled immediately, so most of the function is looking for failing cells in recent columns. Nice.

If I'm looking to see what "Flaky" means, I _might_ stop here, and that's maybe a risk, but I can also dig into the code to figure out how "recent" works or if I think it's broken.

chases2 (approved):
> /cc @chases2
> 
> I am concerned with doing this sort of thing, because people tend to be more fluent in English than golang. Thus minds prefer reading the English. But only the golang determines how the program actually behaves. We need people maintaining this code to read the golang.

Reading the golang is made easier when there's some indication of what the function or author was trying to do. Otherwise, it's tough to judge if code is overengineered, or not sufficient, or bugged.

pkg/summarizer/summary.go
fejta: The thing is, I do not want anyone maintaining this code who "stops there." Maintainers need to read the code, and documentation makes it easier for people to not do this.

PR #469, opened 2021-05-07
Reduce reprocessing time.

fejta:
Currently we reprocess all results that completed within the last 12 hours.
This wastes resources and causes undesirable side effects:
- Continually download the same data for 12 hours
- Large grids may get truncated and only partially process the last 12 hours
  This results in testgrid appearing up to 12 hours stale

We do this because there's a race between uploading the finished.json and
junit files getting uploaded (these are done in parallel). Other systems
can have similiar types of races.

We now mitigate against this by reprocessing all columns at least as recent
as the oldest running column, and we correctly distinguish between running
and completed jobs. So this is no longer necessary.

We protect against jobs stuck running by timing them out after 72 hours.

/cc @michelle192837


fejta:
Passing tests now

michelle192837 (approved):

fejta:
just a rebase so lgtming again

PR #471, opened 2021-05-07
Limit config.proto fields to those we support.

fejta:
There are a large number of fields that non-Google testgrid users do not have access to.
This is a combination of fields which are irrelevant to the prow-based updater,
as well as potentially relevant fields but which have not implemented internally.

Simplify the external-facing proto by limiting the test group config to the set of fields that
are relevant externally

/assign @chases2 @michelle192837 
for your thoughts
/hold


e-blackwelder (commented):
/lgtm

chases2:
I think it might be worth it to keep the Validate() logic together to make imports and error-checking easier.
Let's talk about this in a bit more detail later.

michelle192837 (approved):
Aside from existing comments, I think this looks good to me.

chases2:
I don't think we should merge this change.

Moving validation logic to different locations and repositories makes ensuring that your config is correct for your environment tricky. I much prefer "Here's the validation code, get it from GitHub" over "The definition of valid changes depending on where your TestGrid is"

I would support trimming away validation that isn't used _anywhere_, but this is too sweeping a change. I don't see the benefits as worth it.

fejta:
We should not use the same validation logic internally and externally, as many parts of the internal validation logic are inaccessible and irrelevant externally.

chases2:
Initially, I thought that having a single, correct idea of a valid config was going to be important as we migrate, since having an invalid config can, in general, cause serious disruptions and I wanted to avoid that altogether.

I'm coming around to the idea that the diff between internal and external isn't the core, important parts of the config that can cause outages if they're wrong, but are largely deprecated, finicky, or single-purpose configurations left over from the migration. It's worth nothing that this could cause issues for specific users, but I now think the simplification is worth the extra engineering for them.

Please rebase and submit this change when you get the chance. Thanks for your patience!

fejta:
I'm fine finishing the migration first before worrying about this PR

Issue #478, opened 2021-05-20
simplify readColumns, make single build errors not terminal


fejta:
https://github.com/GoogleCloudPlatform/testgrid/blob/1763aa7a7a5dcd1aa34fd713a64ec2c52db558c3/pkg/updater/read.go#L87-L88 

readColumns is more complicated than necessary because it is trying to process columns in parallel while also preserving their order as well as stop as soon as any fails.

Many of these restrictions no longer apply:
1) We now sort columns at the end, so we do not need to keep their order
2) We should try to process all the requested columns before giving up, and only returning an error if all the columns have an error (aka if we read 18 columns and fail to read 2, add all 20, noting the errors in the two).

PR #480, opened 2021-05-21
Refactor readColumns to only error if all columns fail.

fejta:
Currently they'll stop after the first error. Instead inject a failing overall row into these columns
and continue processing.

ref https://github.com/GoogleCloudPlatform/testgrid/issues/478


fejta:
(passing tests now)

michelle192837 (approved):

chases2 (approved):

PR #483, opened 2021-05-22
Allow cells to collect capture associated issues.

fejta:
Preserve this data across runs


pkg/updater/updater.go
```diff
@@ -608,6 +610,21 @@ func constructGrid(log logrus.FieldLogger, group *configpb.TestGroup, cols []Inf
...
+		issues := make(map[string]bool, len(row.Issues))
+		for _, i := range row.Issues {
+			issues[i] = true
+		}
+		row.Issues = make([]string, 0, len(issues))
+		for i := range issues {
+			row.Issues = append(row.Issues, i)
+		}
+		sort.SliceStable(row.Issues, func(i, j int) bool {
+			return !sortorder.NaturalLess(row.Issues[i], row.Issues[j])
```
michelle192837, line 624: Quick comment on this sort (this is sorting largest issue IDs first, right?) would be nice.

fejta: Done

chases2 (approved):

PR #486, opened 2021-05-25
Add new link templates to use when diffing between columns

mike-michalak:


michelle192837:
/ok-to-test

(I expect this to fail and tell you to run a command to generate the .pb.go files btw, but it should succeed on the run after that.)

michelle192837:
Yup, needs to generate the .pb.go files too ^^ The test link above notes the error:

`ERROR: protos changed. Run bazel run //hack:update-protos`

Should succeed after that.

michelle192837 (approved):

Issue #33, opened 2019-10-14
verify-boilerplate enforces a different boilerplate from the README


chases2:
`//hack:verify_boilerplate` is running the `@io_k8s_repo_infra` boilerplate project, which _requires_ the following boilerplate as a header to go files:

```
/*
Copyright 2019 The Kubernetes Authors.

Licensed under the Apache License, Version 2.0 (the "License");
...
```

The [README](https://github.com/GoogleCloudPlatform/testgrid/tree/549087304d1a9be6c86de41557c4348878a17fe7#source-code-headers) specifies a different boilerplate header. Specifically

```
Copyright 2019 the TestGrid Authors.

Licensed under the Apache License, Version 2.0 (the "License");
...
```

Instead, these should match.

chases2:
Presubmits can now tolerate the README headers: https://github.com/GoogleCloudPlatform/testgrid/tree/master/util/metrics

Not sure when this was fixed, but it's not an issue anymore.

PR #490, opened 2021-05-27
Initial API Implementation

chases2:
API is deployable, but only implements the first endpoint (ListDashboardGroups).

Ready for review, but should not merge until https://github.com/GoogleCloudPlatform/testgrid/pull/488 merges and this is rebased onto it.
/hold

/lint


pkg/api/router.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"context"
+	"path"
+
+	v1 "github.com/GoogleCloudPlatform/testgrid/pkg/api/v1"
+	"github.com/GoogleCloudPlatform/testgrid/util/gcs"
+	"github.com/gorilla/mux"
+)
+
+type RouterOptions struct {
```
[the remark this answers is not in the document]
chases2: Fixed

chases2:
Draft is merged and rebase has been done.
/hold cancel

pkg/api/v1/json.go
```diff
@@ -0,0 +1,34 @@
+/*
+Copyright 2021 The Kubernetes Authors.
```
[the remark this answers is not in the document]
chases2: Unless https://github.com/GoogleCloudPlatform/testgrid/issues/33 has been resolved, doing this will cause these files to fail presubmits.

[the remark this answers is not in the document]
michelle192837: I think we're okay on that front, we have some files using it already (ex: new code like https://github.com/GoogleCloudPlatform/testgrid/tree/master/util/metrics)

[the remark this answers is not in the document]
chases2: Interesting. Fixed here and closed the linked issue.

pkg/api/router.go
```diff
@@ -0,0 +1,54 @@
...
+
+package api
+
+import (
+	"context"
+	"path"
+
+	v1 "github.com/GoogleCloudPlatform/testgrid/pkg/api/v1"
+	"github.com/GoogleCloudPlatform/testgrid/util/gcs"
+	"github.com/gorilla/mux"
```
fejta, line 25: What is this providing over the built-in muxing packages?

chases2: Subrouters aren't supported (to my knowledge) in the built-in Router objects: https://github.com/GoogleCloudPlatform/testgrid/pull/490/files#diff-ce9544cf2f046068d8443822bbe9e602876c44adb5e55e68381ff1ce4e69283cR44-R45

It's a pretty minor quality-of-life change, though; this could be architected to not use the gorilla/mux package. Do you think that would be better?

pkg/api/v1/json.go
```diff
@@ -0,0 +1,34 @@
...
+
+package v1
+
+import (
+	"encoding/json"
+	"net/http"
+)
+
+// WriteJSON will write obj to w as JSON, or will write the JSON marshalling error
+func WriteJSON(w http.ResponseWriter, obj interface{}) {
```
fejta, line 25: This is only called once in one function. How about you just inline it? Or at least put it in the same file where it is used. Or at least stop exporting it to everyone.

IMO we should try and minimally pollute:
* The exported package API
* The total package API
*  The directory filename space.

cmd/api/main.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func main() {
+	log := logrus.WithField("component", "api")
+	opt := gatherOptions()
+
+	log.WithField("port", opt.port).Info("Listening...")
+	router, err := api.GetRouter(opt.router)
+	if err != nil {
+		log.WithField("router-options", opt.router).WithError(err).Fatal("Can't create router")
```
fejta, line 54: doesn't really matter but I think in general log.WithError() tends to come before fields

cmd/api/main.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	if err := http.ListenAndServe(fmt.Sprintf(":%s", opt.port), router); err != nil {
+		log.WithError(err).Error("HTTP Server Error")
```
fejta, line 58: I would expect this to be fatal so that it exits non-zero

pkg/api/router.go
```diff
@@ -0,0 +1,54 @@
...
+// GetRouter returns an http router that serves TestGrid's API
+// It also instantiates necessary caching and i/o objects
+func GetRouter(options RouterOptions) (*mux.Router, error) {
+	r := mux.NewRouter()
+	storageClient, err := gcs.ClientWithCreds(context.Background(), options.GcsCredentials)
+	if err != nil {
+		return nil, err
+	}
+
+	v1Infix := "/api/v1"
```
fejta, line 44: const

pkg/api/router.go
```diff
@@ -0,0 +1,54 @@
...
+	sub1 := r.PathPrefix(v1Infix).Subrouter()
+	s := v1.Server{
+		Client:        gcs.NewClient(storageClient),
+		Host:          path.Join(options.Hostname, v1Infix),
+		DefaultBucket: options.HomeBucket,
+	}
+	sub1.HandleFunc("/dashboardgroups", s.ListDashboardGroups).Methods("GET")
```
fejta, line 51: shouldn't this be dashboardGroups?

pkg/api/v1/config.go
```diff
@@ -0,0 +1,81 @@
...
+	if s.DefaultBucket != "" {
+		return gcs.NewPath(fmt.Sprintf("%s/%s", s.DefaultBucket, "config"))
+	}
+	return nil, errors.New("no testgrid scope")
+}
+
+// passQueryParameters returns only the query parameters in the request that need to be passed through to links
+func passQueryParameters(r *http.Request) string {
+	scope := r.URL.Query().Get("scope")
+	if scope != "" {
```
fejta, line 50: You're not using scope outside of this if statement, so combine these lines
```
if scope := r.URL.Query().Get("scope"); scope != "" {
  return ...
}
```

pkg/api/router.go
fejta: In general I think N deps is easier to reason about and maintain than N+1 deps. So the cost is non-zero, but the benefits can outweigh those costs.

Probably fine here, not totally clear on why this is necessary/helpful.

pkg/api/router.go
chases2: On reflection, I think both of these are wrong. It's tough to find specific recommendations for what case resources and their endpoints should be in, but `hypenated-case` seems better than both and resolves the "is it case sensitive" problems I'm trying to avoid.

https://cloud.google.com/apis/design/resource_names
https://blog.restcase.com/5-basic-rest-api-design-guidelines/

pkg/api/v1/json.go
chases2: I put it in its own file since it seemed like this helper would be useful everywhere: whether we're writing a config, a summary, or whatever, this helper seemed a logical extension to `http.Error`.

I definitely want to use it all over this package, but I can do that without exporting it to everyone.

fejta: That seems like a good optimization to make once foo.go, bar.go and frobber.go are all calling this method.

Right now though, this is only used by foo.go and thus seems to belong in foo.go

fejta:
Can you please rebase this?

pkg/api/v1/config.go
```diff
@@ -0,0 +1,80 @@
...
+
+package v1
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+
+	"github.com/GoogleCloudPlatform/testgrid/config"
+	pb "github.com/GoogleCloudPlatform/testgrid/pb/api/v1"
```
fejta, line 25: apipb tends to be the more common pattern

fejta (approved):
/hold

pkg/api/v1/config.go
chases2: Fixed
/hold cancel

pkg/api/router.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+// RouterOptions are the options needed to GetRouter
+type RouterOptions struct {
+	GcsCredentials string
+	Hostname       string
+	HomeBucket     string
+}
+
+const v1Infix = "/api/v1"
```
fejta, line 35: Please move inside GetRouter where it is used (aka keep it scoped to `GetRouter()`)

fejta (approved):
/hold

fejta (approved):

fejta (approved):

PR #495, opened 2021-05-28
Allow percent difference when comparing states.

michelle192837:
This allows for small differences when comparing states.

Also adds a couple of other tweaks (short vs verbose reports, provide a URL to get tab-links-from-test-groups for diffed groups, update row and column extraction to be less exact given known differences in old vs new updater).


michelle192837:
This passes now, so ready for review!

fejta (approved):

PR #502, opened 2021-06-03
Add log-only metric reporter.

michelle192837:


fejta:
I think you need to rerun update-bazel and/or `git add -A`

fejta (commented):
Would I be annoying to ask that we use it somewhere? 

For example: https://github.com/GoogleCloudPlatform/testgrid/blob/14ec4b9a9e7aa10f59072d783a9b9d37bb05b4dd/cmd/updater/main.go#L137-L138

michelle192837:
> Would I be annoying to ask that we use it somewhere?
> 
> For example:
> 
> https://github.com/GoogleCloudPlatform/testgrid/blob/14ec4b9a9e7aa10f59072d783a9b9d37bb05b4dd/cmd/updater/main.go#L137-L143

Np! Does this work?

cmd/updater/main.go
```diff
@@ -136,9 +137,11 @@ func main() {
 	groupUpdater := updater.GCS(opt.groupTimeout, opt.buildTimeout, opt.buildConcurrency, opt.confirm, updater.SortStarted)
 	updateOnce := func() {
 		start := time.Now()
+		cycle := metrics.NewLogInt64("cycle_duration", "How long an update cycle took, in seconds.", logrus.New())
```
fejta, line 140: Shouldn't we just create this once, before updateOnce definition? (so L137 or so?)

michelle192837: Yup, thanks for catching! 

Also manually verified that this will successfully log when the Updater is run.

fejta (approved):

PR #508, opened 2021-06-09
Fix lint issues in the repo.

michelle192837:
/lint

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -35,7 +35,7 @@ import (
 )
 
 // MissingFieldError is an error that includes the missing root field.
-// Entities that contain no children should use a ConfigError, so they can point to the empty Entity
+// Entities that contain no children should use a ValidationError, so they can point to the empty Entity
 type MissingFieldError struct {
 	Field string
 }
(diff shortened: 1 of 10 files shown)


config/config.go
```diff
@@ -64,14 +64,14 @@ func (e MissingEntityError) Error() string {
 	return fmt.Sprintf("could not find the referenced (%s) %s", e.Entity, e.Name)
 }
 
-// ConfigError is an error for invalid configuration that includes what entity errored.
-type ConfigError struct {
+// Error is an error for invalid configuration that includes what entity errored.
+type Error struct {
```
chases2, line 68: I think "ConfigError" or "EntityError" is better. Easy to typo or confuse `error` with `Error`

chases2 (commented):
Thanks for this! Only one comment

config/config.go
fejta: Issue is that this gets written as `config.ConfigError` outside the package, which stutters, and why the linter will complain about it.

chases2: That's a good point. "ValidationError"?

michelle192837: ValidationError SGTM, updated!

chases2 (approved):

PR #509, opened 2021-06-10
Optionally print all config fields and # of uses.

michelle192837:
To run:
`bazel run //config/print:print -- -path=/tmp/testgrid-states/config -print-field-use=true`

### DIFF
--- a/config/print/main.go
+++ b/config/print/main.go
@@ -30,8 +30,9 @@ import (
 )
 
 type options struct {
-	configPath string
-	creds      string
+	configPath    string
+	creds         string
+	printFieldUse bool
 }
 
 func gatherOptions() options {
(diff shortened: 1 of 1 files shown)


config/print/main.go
```diff
@@ -66,4 +68,58 @@ func main() {
...
+	type Dashboard struct {
+		DashboardTabs []map[string]interface{} `json:"dashboard_tab"`
+	}
+	type Config struct {
+		TestGroups      []map[string]interface{} `json:"test_groups"`
+		Dashboards      []Dashboard              `json:"dashboards"`
+		DashboardGroups []map[string]interface{} `json:"dashboard_groups"`
+	}
+	var output Config
+	err := json.Unmarshal(b, &output)
```
chases2, line 90: You can avoid a marshal/unmarshal by using the "cfg" on line 61 instead of the JSON byte array.

chases2 (commented):
A "count fields" function is great for debugging. Thanks!

config/print/main.go
michelle192837: I'm not sure I can avoid it, since this weird workaround needs to marshal into a particular struct to make counting the fields possible? (Might be missing something though)

chases2: Ah, I see. You could get the field names with the reflect library to avoid the serialize/deserialize, but I can see the argument that this is also a clear and good way of doing this.

/approve
/lgtm

PR #511, opened 2021-06-11
Add basic success/error metrics for components.

michelle192837:
Manually tested that these will log when the components are run.


pkg/updater/updater.go
```diff
@@ -132,36 +186,7 @@ func Update(parent context.Context, client gcs.ConditionalClient, configPath gcs
...
-					}
-					log.Debug("Acquired update lock")
-				}
-				if err := updateGroup(ctx, log, client, &tg, *tgp); err != nil {
-					log.WithError(err).Error("Error updating group")
-				}
-				// run the garbage collector after each group to minimize
-				// extraneous memory usage.
-				runtime.GC()
+				update(ctx, client, mets, log, &tg, configPath, gridPrefix, updateGroup, write, generations)
```
fejta, line 189: This seems a lot cleaner than futzing around with `var errored bool`

```golang
if err := update(...); err != nil {
  mets.Errors.Add(1, "updater")
} else {
  mets.Successes.Add(1, "updater")
}
```

pkg/updater/updater.go
```diff
@@ -109,8 +115,56 @@ func lockGroup(ctx context.Context, client gcs.ConditionalClient, path gcs.Path,
 	return gcs.Touch(ctx, client, path, generation, buf)
 }
 
+func update(ctx context.Context, client gcs.ConditionalClient, mets Metrics, log logrus.FieldLogger, tg *configpb.TestGroup, configPath gcs.Path, gridPrefix string, updateGroup GroupUpdater, write bool, generations map[string]int64) {
+	var errored bool
+	defer func() {
+		if errored {
+			if mets.Errors != nil {
+				mets.Errors.Add(1, "updater")
```
fejta, line 123: Seems kind of strange to make the caller remember to add an updater string. Seems like this should be embedded into either the mets or Errors structure.

pkg/updater/updater_test.go
```diff
@@ -229,6 +229,7 @@ func TestUpdate(t *testing.T) {
 			err := Update(
 				ctx,
 				client,
+				Metrics{},
```
fejta, line 232: Can we unit test this new scenario? AKA successes should increment the success counter, errors should increment the error counter.

pkg/updater/updater.go
```diff
@@ -109,8 +115,56 @@ func lockGroup(ctx context.Context, client gcs.ConditionalClient, path gcs.Path,
...
+			if mets.Successes != nil {
+				mets.Successes.Add(1, "updater")
+			}
+		}
+	}()
+	log = log.WithField("group", tg.Name)
+	log.Debug("Starting update")
+	tgp, err := testGroupPath(configPath, gridPrefix, tg.Name)
+	if err != nil {
+		log.WithError(err).Error("Bad path")
```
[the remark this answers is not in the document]
michelle192837: Fair enough, but it's in the original code so leaving it in for now.

pkg/updater/updater.go
```diff
@@ -109,8 +115,56 @@ func lockGroup(ctx context.Context, client gcs.ConditionalClient, path gcs.Path,
...
+		return
+	}
+	if write && generations != nil {
+		if err := lockGroup(ctx, client, *tgp, generations[tg.Name]); err != nil {
+			var ok bool
+			switch ee := err.(type) {
+			case *googleapi.Error:
+				if ee.Code == http.StatusPreconditionFailed {
+					ok = true
+					log.Debug("Lost the lock race")
```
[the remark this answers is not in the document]
michelle192837: Good point, adding a TODO for now

fejta (approved):

PR #516, opened 2021-06-22
Test grid dynamic email list.

yairnee:
The idea is that the user will be able to set a dynamic email list for alert.


pkg/updater/gcs.go
```diff
@@ -386,6 +388,13 @@ func convertResult(log logrus.FieldLogger, nameCfg nameConfig, id string, header
 		out.Column.Extra = append(out.Column.Extra, val)
 	}
 
+	emailAddresses, ok := result.finished.Finished.Metadata[EmailListKey].([]string)
```
yairnee, line 391: Maybe  we should split this.
If we can't find Metadata[EmailListKey] set an empty list.
If the value is not '[]string' we should throw warning.
WDYT?

pkg/updater/gcs_test.go
```diff
@@ -335,6 +335,30 @@ func TestConvertResult(t *testing.T) {
 				},
 			},
 		},
+		{
```
michelle192837, line 338: Please add some more test cases (for multiple emails and invalid email lists)

michelle192837 (changes requested):

pkg/updater/gcs.go
yairnee: We should log an error if invalid email list is given? I didn't saw any handling of invalid input (or handling errors) in the code.

pkg/updater/updater.go
```diff
@@ -880,11 +884,11 @@ func alertRow(cols []*statepb.Column, row *statepb.Row, failuresToOpen, passesTo
 		latestID = row.CellIds[latestFailIdx]
 	}
 	msg := row.Messages[latestFailIdx]
-	return alertInfo(totalFailures, msg, id, latestID, firstFail, latestFail, latestPass)
+	return alertInfo(totalFailures, msg, id, latestID, firstFail, latestFail, latestPass, latestColumn)
```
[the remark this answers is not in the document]
yairnee: In our case the idea is to send emails until it's succeeded. 
If i'll use firstFail I'll keep sending emails to the user, right? 
Or because it's part of the "old_dasboard" we will not resend emails about it.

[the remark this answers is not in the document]
michelle192837: Using firstFail vs. latestColumn doesn't actually matter for how often the mails get sent, that's controlled by user configuration. (It should only send an email when new failing rows are noticed, _or_ periodically until the tab passes if the user's configured it to do so).

Given that, I think firstFail might be more appropriate here; it matches with the 'start' of the failure.

pkg/updater/gcs.go
michelle192837: I think logging an error is fine here, yeah. (If it doesn't cast correctly to a []string then we should note that somehow).

pkg/updater/updater.go
```diff
@@ -895,7 +899,15 @@ func alertInfo(failures int32, msg, cellID, latestCellID string, fail, latestFai
 		FailureMessage:    msg,
 		PassTime:          stamp(pass),
 		PassBuildId:       buildID(pass),
+		EmailAddresses:    emailAddresses(latestColumn),
+	}
+}
+
+func emailAddresses(col *statepb.Column) []string {
```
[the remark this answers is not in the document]
michelle192837: I believe it'll work with the pointer even if the pointer's nil, but this is also fine!

pkg/updater/updater_test.go
```diff
@@ -3128,6 +3128,62 @@ func TestAppendColumn(t *testing.T) {
 	}
 }
 
+func TestDynamicEmails(t *testing.T) {
+	columnWithEmails := statepb.Column{Build: "columnWithEmail", Started: 100 - float64(0), EmailAddresses: []string{"email1@", "email2@"}}
+	anotherColumnWithEmails := statepb.Column{Build: "anotherColumnWithEmails", Started: 100 - float64(1), EmailAddresses: []string{"email3@", "email2@"}}
+	columnWithoutEmails := statepb.Column{Build: "columnWithoutEmails", Started: 100 - float64(2)}
+	cases := []struct {
```
michelle192837, line 3134: Think 1 more test case would be a good idea (2 columns have dynamic emails, and the correct emails are chosen).

michelle192837 (commented):
Thanks! This looks good once the test case comment I just added is addressed. ^^

pkg/updater/updater.go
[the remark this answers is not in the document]
yairnee: But then it's not supposed to be latestFail and not firstFail? 
Because we want to alert about the latest failure happens

michelle192837 (approved):

PR #519, opened 2021-06-26
Refactor Opener interface to return attributes.

fejta:
This allows the consumer of the open file to know various metadata, such as when it was last modified and/or
the object generation. This can help minimize conflicts by minimizing the time between when we discover the
generation of a downloaded object and when we attempt to update it, assuming the remote generation is unchanged.


michelle192837 (approved):

fejta:
Sorry think this one was out of date or something, can you rereview?

michelle192837 (approved):

PR #522, opened 2021-07-03
Refactor updater to continually send groups to be updated.

fejta:
Currently updater will wait until all groups have been updated once before updating them a second time.
Instead groups are put into a priority queue based on how recently they were updated.
As soon as an item is sent for an update, its moved to the back of the queue.

This helps mitigate the situation where 99% of the groups have been updated but the last one is taking a long time.
Instead of waiting for this one to finish, the other groupConcurrency workers will being updating the other groups
while this one is finishing.

### DIFF
--- a/config/queue_test.go
+++ b/config/queue_test.go
@@ -1,0 +1,689 @@
+/*
+Copyright 2021 The TestGrid Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package config
+
+import (
+	"container/heap"
+	"context"
+	"sync"
+	"testing"
+	"time"
+
+	configpb "github.com/GoogleCloudPlatform/testgrid/pb/config"
+	"github.com/google/go-cmp/cmp"
+	"google.golang.org/protobuf/testing/protocmp"
+)
+
+func TestInit(t *testing.T) {
+	now := time.Now()
+	cases := []struct {
+		name   string
+		q      *TestGroupQueue
+		groups []*configpb.TestGroup
+		when   time.Time
+
+		next []*configpb.TestGroup
+	}{
+		{
+			name: "add",
+			q:    &TestGroupQueue{},
+			groups: []*configpb.TestGroup{
+				{
+					Name: "hi",
+				},
+			},
+			when: now,
+
+			next: []*configpb.TestGroup{
+				{
+					Name: "hi",
+				},
+			},
+		},
+		{
+			name: "remove",
+			q: func() *TestGroupQueue {
+				var q TestGroupQueue
+				q.Init([]*configpb.TestGroup{
+					{
+						Name: "drop",
+					},
+					{
+						Name: "keep",
+					},
+				}, now)
+				return &q
+			}(),
+			groups: []*configpb.TestGroup{
+				{
+					Name: "keep",
+				},
+				{
+					Name: "add",
+				},
+			},
+			when: now.Add(-time.Minute),
+
+			next: []*configpb.TestGroup{
+				{
+					Name: "add",
+				},
+				{
+					Name: "keep",
+				},
+			},
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			tc.q.Init(tc.groups, tc.when)
+
+			var got []*configpb.TestGroup
+			for range tc.next {
+				got = append(got, heap.Pop(&tc.q.queue).(*item).tg)
+			}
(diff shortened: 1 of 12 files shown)


config/queue_test.go
```diff
@@ -0,0 +1,686 @@
...
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+
+			ctx, channel, get := tc.receivers(ctx)
+			if err := tc.q.Send(ctx, channel, tc.freq); err != ctx.Err() {
```
chases2, line 582: From reading the code, I would have expected "multi" and "single" to return no error.

It seems like the second `<-ctx.Done()` check on ll. 183-185 are there so that this returns with the "correct" error instead of no error. Is that necessary? It seems better to just have an `expectError` in your cases and have these two modes terminate differently.

pkg/updater/updater.go
```diff
@@ -131,97 +117,185 @@ func lockGroup(ctx context.Context, client gcs.ConditionalClient, path gcs.Path,
...
-	groups := make(chan configpb.TestGroup)
+	var lock sync.Mutex
 	var wg sync.WaitGroup
+	wg.Add(groupConcurrency)
 	defer wg.Wait()
-	defer close(groups)
-
-	var generations map[string]int64
-
+	channel := make(chan *configpb.TestGroup) // TODO(fejta): pass into this function
```
chases2, line 231: `// TODO(fejta): pass into this function` Why do we want this?

pkg/updater/updater.go
```diff
@@ -131,97 +117,185 @@ func lockGroup(ctx context.Context, client gcs.ConditionalClient, path gcs.Path,
...
+		ticker := time.NewTicker(time.Minute)
+		for {
+			now := time.Now()
+			depth, next, when := q.Status()
+			log := log.WithField("depth", depth)
+			if next != nil {
+				log = log.WithField("next", next.Name)
+			}
+			if delay := now.Sub(when).Round(time.Second); delay > 0 {
+				log = log.WithField("delay", delay)
```
chases2, line 276: Delay seems like it might be a useful metric to track instead of cycle duration. What do you think?

chases2 (commented):
This is a great improvement! A priority queue is a great way to deal with these updates of various lengths. Very solid code, but it's doing a lot and I have a few questions and comments about it.

config/queue_test.go
fejta: 1) Yes, both `select` statements are necessary.

The first `select{}` is to check whether the context expired while we were waiting for the lock.

The second `select{}` is to detect when the context expires while we're waiting for someone to read the channel we want to write to.

2) Yes the expectation is single scenarios return nil and multi scenarios return the contextual error.

AKA if `ctx.Err() == nil` (because we popped all the elements and the queue is empty) then this will `if err != nil { t.Error() }`.

When `frequency != 0` then `Send()` runs forever, at least until we cancel the context we sent it. Note the `if len(got) == 2 { cancel() }` lines. In this situation `ctx.Err() != nil` and this will `want := ctx.Err(); err != want { t.Error() }`

pkg/updater/updater.go
fejta: Done

pkg/updater/updater.go
fejta: Done

pkg/updater/updater.go
```diff
@@ -131,97 +117,185 @@ func lockGroup(ctx context.Context, client gcs.ConditionalClient, path gcs.Path,
...
+		}
+		q.FixAll(updates)
+	}
+	return configGen, generations, nil
 }
 
-// Update performs a single update pass of all all test groups specified by the config.
-func Update(parent context.Context, client gcs.ConditionalClient, mets *Metrics, configPath gcs.Path, gridPrefix string, groupConcurrency int, group string, updateGroup GroupUpdater, write bool) error {
-	defer growMaxUpdateArea()
+// Updates test groups with the specified freq.
```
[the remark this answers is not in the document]
fejta: Done

pkg/updater/updater.go
```diff
@@ -131,97 +117,185 @@ func lockGroup(ctx context.Context, client gcs.ConditionalClient, path gcs.Path,
...
+	return e.Code == http.StatusPreconditionFailed
+}
+
+func update(ctx context.Context, client gcs.ConditionalClient, log logrus.FieldLogger, tg *configpb.TestGroup, tgp gcs.Path, updateGroup GroupUpdater, write bool, gen int64) error {
+	log.Debug("Starting update")
+	if write && gen >= 0 {
+		if attrs, err := lockGroup(ctx, client, tgp, gen); err != nil {
+			if !isPreconditionFailed(err) {
+				return fmt.Errorf("lock: %v", err)
+			} else {
```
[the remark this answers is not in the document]
fejta: Done

config/queue_test.go
```diff
@@ -0,0 +1,686 @@
...
+				{
+					Name: "basic",
+				},
+			},
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			if err := tc.q.Fix(tc.name, tc.when); (err != nil) != tc.err {
```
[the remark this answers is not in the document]
fejta: Done

michelle192837 (approved):
Reviewing since folks are out today.

PR #527, opened 2021-07-09
Refactor metrics interface.

fejta:
Change Val to Values() and make it specific to logging metrics.
Create a reporter which only logs metric values periodically

### DIFF
--- a/util/metrics/log_test.go
+++ b/util/metrics/log_test.go
@@ -17,131 +17,226 @@ limitations under the License.
 package metrics
 
 import (
-	"github.com/sirupsen/logrus"
-	"github.com/sirupsen/logrus/hooks/test"
 	"testing"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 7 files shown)


util/metrics/log_test.go
```diff
@@ -17,131 +17,226 @@ limitations under the License.
...
+				},
+			},
 		},
 		{
-			name:       "not enough fields",
-			n:          3,
-			fieldNames: []string{"user"},
-			fields:     []string{},
-			err:        true,
+			name:   "values",
```
michelle192837, line 79: nit: This is named the same as the last test case

michelle192837 (approved):
/hold

for comment

util/metrics/log_test.go
fejta: Done

fejta:
/hold cancel

Ready for another review

michelle192837 (approved):

Issue #500, opened 2021-06-02
TestInflateDropAppend flaky


fejta:
```
=== RUN   TestInflateDropAppend/recent
time="2021-06-02T23:06:24Z" level=info msg="Wrote grid" bytes=193 cols=3 rows=1 test=recent url="gs://fake/upload/location"
    TestInflateDropAppend/recent: updater_test.go:1677: updateGroup() got unexpected diff (-want +got):
          fake.Uploader{
          	s"gs://fake/upload/location": {
          		Buf: []uint8{
          			0x78, 0x9c, 0xe2, 0x92, 0xe5, 0x62, 0x4f, 0x2e, 0x2d, 0x2a, 0x4a, 0xcd, 0x2b, 0x91, 0x64, 0x68, //  |x....bO.-*J.+.dh|
          			0x88, 0x5c, 0xfa, 0x7e, 0x4e, 0xb9, 0x93, 0x12, 0x83, 0x11, 0x4c, 0x8c, 0x4b, 0x87, 0x8b,       //  |.\.~N.....L.K..|
        - 			0xb7, 0x20, 0xb1, 0xb8,                                                                         // -|. ..|
        + 			0x3b, 0xb1,                                                                                     // +|;.|
          			0x44, 0x21, 0x29, 0xbf, 0x34, 0x2f, 0x25, 0xb1, 0xa8, 0x52, 0x92,                               //  |D!).4/%..R.|
        - 			0xe1, 0x83, 0xda, 0x62, 0xb0, 0x22, 0x96, 0xec, 0xd4, 0xd4, 0x02, 0x23, 0x1e, 0x98, 0x84, 0xb6, // -|...b.".....#....|
        - 			0xa5, 0xa5, 0xa5, 0x90, 0x17, 0x17, 0xbb, 0x7f, 0x59, 0x6a, 0x51, 0x62, 0x4e, 0x8e, 0x10, 0x8c, // -|........YjQbN...|
        - 			0x21, 0xc5, 0xc2, 0xc2, 0xc8, 0xcb, 0xa8, 0xc4, 0xa0, 0xc4, 0xa0, 0x25, 0xe6, 0x54, 0x9a, 0x99, // -|!..........%.T..|
        - 			0x93, 0xa2, 0x50, 0x5c, 0x92, 0x99, 0x93, 0xa3, 0x50, 0x54, 0x9a, 0x97, 0x97, 0x99, 0x97, 0xae, // -|..P\....PT......|
[...]

PR #537, opened 2021-07-10
Lock access to reading the generation.

fejta:
Golang will panic if a map is concurrently read and written.
We guard the map write (L303) but we also need to protect when we read from this map.

Switch to use an RWMutex so that we can concurrently read.


fejta:
Passing tests, flaky test is #500

chases2 (approved):

PR #539, opened 2021-07-13
Fix linting and bad copying in api package

chases2:
"bad copying" == "Don't copy locks" and "Don't copy big protos"

/assign @fejta 
/lint

### DIFF
--- a/pkg/api/v1/config.go
+++ b/pkg/api/v1/config.go
@@ -14,6 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+// Package v1 (api/v1) is the first versioned implementation of the API
 package v1
 
 import (
@@ -76,5 +77,5 @@ func (s Server) ListDashboardGroups(w http.ResponseWriter, r *http.Request) {
 		groups.DashboardGroups = append(groups.DashboardGroups, &rsc)
 	}
 
-	writeJSON(w, groups)
+	writeJSON(w, &groups)
 }
--- a/pkg/api/v1/config_test.go
+++ b/pkg/api/v1/config_test.go
@@ -118,56 +118,88 @@ func TestPassQueryParameters(t *testing.T) {
 
 func TestListDashboardGroups(t *testing.T) {
 	tests := []struct {
-		name         string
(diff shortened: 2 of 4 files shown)


pkg/api/v1/config.go
```diff
@@ -14,6 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+// Package api.v1 is the first versioned implementation of the API
```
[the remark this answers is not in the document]
chases2: Fixed

pkg/api/v1/config_test.go
```diff
@@ -154,21 +155,22 @@ func TestListDashboardGroups(t *testing.T) {
...
-			t.Errorf("Expected OK status, but got %v", response.Code)
-		}
-		if response.Body.String() != test.expectedJSON {
-			t.Errorf("In Body, Expected %s; got %s", test.expectedJSON, response.Body.String())
-		}
+		t.Run(test.name, func(t *testing.T) {
+			server := setupTestServer(t, test.config)
+			request, err := http.NewRequest("GET", "/api/v1/dashboardgroups", nil)
+			if err != nil {
+				t.Fatalf("Can't form dashboardgroup request")
```
fejta, line 162: Include the error message?

`t.Fatalf("whatever: %v", err)`

pkg/api/v1/config_test.go
```diff
@@ -154,21 +155,22 @@ func TestListDashboardGroups(t *testing.T) {
...
+			}
+			response := httptest.NewRecorder()
+			server.ListDashboardGroups(response, request)
+			if response.Code != http.StatusOK {
```
fejta, line 166: Shouldn't we also test non-happy paths?

fejta (approved):
/hold

pkg/api/v1/config_test.go
chases2: Done.

pkg/api/v1/config_test.go
chases2: Done. Expanded this test to include passed-in scopes and trying to read an invalid config.

PR #542, opened 2021-07-13
Reduce concurrency.

fejta:
We are running into evictions because of memory pressure, so reduce the
number of builds and groups we process in parallel.

This, combined with the fact that we now process new groups last means we struggle to process new groups.
https://github.com/GoogleCloudPlatform/testgrid/blob/95cfeb295b8d6e732fe4cc3f8ab74e67ed33ea8e/config/queue.go#L185

/cc @chizhg


chizhg:
/lgtm
/hold
In case Michelle and Sean have comments.

chases2 (approved):

fejta:
/hold cancel

Thanks to you both!

PR #548, opened 2021-07-16
API deployment cluster in canary (only)

chases2:
Adds extra logging when a default scope is used but still can't read

### DIFF
--- a/cluster/canary/api.yaml
+++ b/cluster/canary/api.yaml
@@ -1,0 +1,53 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: testgrid-api
+  namespace: testgrid-canary
+  labels:
+    app: testgrid
+    channel: stable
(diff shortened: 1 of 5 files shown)


michelle192837 (approved):

cluster/canary/api.yaml
```diff
@@ -0,0 +1,53 @@
...
+metadata:
+  annotations:
+    iam.gke.io/gcp-service-account: testgrid-canary-api@k8s-testgrid.iam.gserviceaccount.com
+  name: api
+  namespace: testgrid-canary
+---
+kind: Service
+apiVersion: v1
+metadata:
+  name: api
```
fejta, line 44: this is missing `namespace:`

fejta: https://github.com/GoogleCloudPlatform/testgrid/pull/552

PR #550, opened 2021-07-17
Implement GetDashboardGroup

chases2:
xref #549 

- Refactors testing to use a router instead of server functions directly so test cases are more intuitive
- Deduplicates reused code between the two API functions


e-blackwelder (commented):
/lgtm
/hold

You've got some test failures... Go can't import "github.com/gorilla/mux" or "github.com/GoogleCloudPlatform/testgrid/pb/config"

chases2:
Added the boilerplate & bazel imports; bazel tests are now passing locally.

PR #553, opened 2021-07-20
Use buffered channels instead of concurrency.

fejta:
Currently we start many go threads to read artifacts and builds in parallel. This causes a lot of
memory pressure. So instead switch to channels with a 1 item buffer. This allows each go routine
to continue working on the second item as soon as it finishes the first one. Things will only block
if the second item is completed before the first item is read.

Assuming everything is processing quickly this should result in similar performance but with better
performance when things slow down (memory consumed for 1-2 items instead of N).

Also: reject junit files more than 100MB, failing these builds with a note about the error.

### DIFF
--- a/pkg/updater/read_test.go
+++ b/pkg/updater/read_test.go
@@ -22,7 +22,6 @@ import (
 	"encoding/xml"
 	"errors"
 	"net/url"
-	"reflect"
 	"sort"
 	"sync"
 	"testing"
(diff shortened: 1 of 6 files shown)


pkg/updater/read_test.go
```diff
@@ -1617,11 +1616,21 @@ func TestReadSuites(t *testing.T) {
 			err: true,
 		},
 		{
-			name: "suites error returns error",
+			name: "suites error conatins error",
```
michelle192837, line 1619: nit:contains

michelle192837 (approved):

michelle192837:
Oh whoops, late

/hold

pkg/updater/read_test.go
fejta: Done

fejta:
ref https://github.com/GoogleCloudPlatform/testgrid/pull/553

PR #561, opened 2021-07-20
Check preconditions when reading.

fejta:
The official storage client does this incorrectly: it uses the XML api for reads but sends ifGenerationMatch
parameters (for the json api) rather than the required x-goog-if-generation-match headers. So instead check
this manually when the response comes back.

Also:
* fix some related issues (cannot use the testgroup's conditions to read objects, similar thing for reading
the config).
* Only fix the update time if the group already exists and has a written state
* Only fix the heap when the update time changes.

### DIFF
--- a/cmd/updater/main.go
+++ b/cmd/updater/main.go
@@ -155,7 +155,7 @@ func main() {
 		"build": opt.buildConcurrency,
 	}).Info("Configured concurrency")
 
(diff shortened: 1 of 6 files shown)


chases2 (approved):

fejta:
ref https://github.com/googleapis/google-cloud-go/issues/4470

danielduhh:
@fejta the fix was rolled into the[ 1.16.1 go release ](https://github.com/googleapis/google-cloud-go/releases/tag/storage%2Fv1.16.1)

fejta:
Thanks for the update!

PR #564, opened 2021-07-21
Synthetic columns are not readColumns errors.

fejta:
Initially the updater aborted the test group when it encountered a problematic build.
Next it would just ignore the column.
Today it emits a synthetic column that gives information about what parts of the build
are problematic.

This obviates the need to have readColumns() return an error indicating whether or not it found
any error-free columns.

Today so long as 1 column is free of errors we return a nil error (even if 9 others have problems).
This changes it so that we'll also return nil error it that last column also contains an error.
Since we always return a nil error, remove this from the function signature.

### DIFF
--- a/pkg/updater/read.go
+++ b/pkg/updater/read.go
@@ -63,14 +63,15 @@ func gcsColumnReader(client gcs.Client, buildTimeout time.Duration, concurrency 
 		}
 		log.WithField("total", len(builds)).Debug("Listed builds")
 
-		return readColumns(ctx, client, log, tg, builds, stop, buildTimeout, receivers)
+		readColumns(ctx, client, log, tg, builds, stop, buildTimeout, receivers)
+		return nil
 	}
 }
 
 // readColumns will list, download and process builds into inflatedColumns.
-func readColumns(ctx context.Context, client gcs.Downloader, log logrus.FieldLogger, group *configpb.TestGroup, builds []gcs.Build, stop time.Time, buildTimeout time.Duration, receivers chan<- InflatedColumn) error {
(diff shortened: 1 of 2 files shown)


michelle192837 (approved):

pkg/updater/read.go
```diff
@@ -63,14 +63,15 @@ func gcsColumnReader(client gcs.Client, buildTimeout time.Duration, concurrency
...
-		return readColumns(ctx, client, log, tg, builds, stop, buildTimeout, receivers)
+		readColumns(ctx, client, log, tg, builds, stop, buildTimeout, receivers)
+		return nil
 	}
 }
 
 // readColumns will list, download and process builds into inflatedColumns.
-func readColumns(ctx context.Context, client gcs.Downloader, log logrus.FieldLogger, group *configpb.TestGroup, builds []gcs.Build, stop time.Time, buildTimeout time.Duration, receivers chan<- InflatedColumn) error {
+func readColumns(ctx context.Context, client gcs.Downloader, log logrus.FieldLogger, group *configpb.TestGroup, builds []gcs.Build, stop time.Time, buildTimeout time.Duration, receivers chan<- InflatedColumn) {
 	if len(builds) == 0 {
```
chases2, line 73: Note: this does make the case where len(builds) == 0 go from "gcsColumReader returns an error" to "silently do nothing". I think the latter is probably better behavior; feel free to unhold if you agree.

chases2 (approved):
/hold for comment

pkg/updater/read.go
fejta: It wouldn't return an error in this situation previously though (it returns nil when there are no builds)

PR #576, opened 2021-07-28
Fix a few bugs in config validation.

michelle192837:
Fixing up a few bugs I noticed while evaluating another config yesterday:
- tabular_name_regex is counting subexpressions wrong due to how regexp.SubexpNames() works
- short text should be > not >= and count runes (5 emoji is fine to display), not characters
- an empty gcs_prefix is valid if result_source is specified

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -25,6 +25,7 @@ import (
 	"os"
 	"regexp"
 	"strings"
+	"unicode/utf8"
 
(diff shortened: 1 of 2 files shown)


michelle192837:
Also our config test may need a small rewrite to be in line with our style, but not doing that this PR.

chases2 (approved):

PR #584, opened 2021-07-29
Ignore columns with no cells.

michelle192837:
ETA: Instead of just throwing away empty columns, this groups all empty columns together, keeping the hint of the latest empty column. Thanks Erick for the suggestion!

This ignores empty columns instead of processing them into the grid. (This will not progress the latest 'hint' for the grid past the last non-empty column, so it's possible to repeat work, but in my estimation this should not be a significant issue. If it is, swap to an artificial 'last processed' column that we keep at the beginning to make sure the hint progresses).

### DIFF
--- a/pkg/updater/updater.go
+++ b/pkg/updater/updater.go
@@ -533,6 +533,11 @@ func InflateDropAppend(ctx context.Context, alog logrus.FieldLogger, client gcs.
 	case <-ctx.Done():
 		return false, fmt.Errorf("first column: %w", ctx.Err())
 	case col := <-newCols:
+		if len(col.Cells) == 0 {
+			// Group all empty columns together by setting build/name empty.
+			col.Column.Build = ""
+			col.Column.Name = ""
+		}
 		cols = append(cols, col)
 	case err := <-ec:
 		if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/updater/updater.go
```diff
@@ -533,7 +533,9 @@ func InflateDropAppend(ctx context.Context, alog logrus.FieldLogger, client gcs.
 	case <-ctx.Done():
 		return false, fmt.Errorf("first column: %w", ctx.Err())
 	case col := <-newCols:
-		cols = append(cols, col)
+		if len(col.Cells) != 0 {
```
fejta, line 536: A concern I have with this approach is that this will prevent us from updating the hint....

What would you think about keeping around the most recently received empty column to mitigate this?

fejta: OK I guess you specifically call that out in the PR description but I've been fighting a lot of hint-related issues so I'm a bit concerned.

fejta (approved):
/hold

pkg/updater/updater.go
fejta: PS: I'd do

```golang
if len(col.Cells) == 0 {
  continue
}
cols = append(cols, col)
```

fejta (approved):

pkg/updater/updater.go
michelle192837: Discussed offline, and implemented the "group empty columns into one empty column" idea.

PR #599, opened 2021-07-31
Handle unused ctx in test column reader.

michelle192837:


pkg/updater/updater_test.go
```diff
@@ -1059,6 +1059,40 @@ func TestInflateDropAppend(t *testing.T) {
 	now := time.Now().Unix()
 	uploadPath := newPathOrDie("gs://fake/upload/location")
 	defaultTimeout := 5 * time.Minute
+	// a simple ColumnReader that parses fakeBuilds
+	fakeColReader := func(builds []fakeBuild) ColumnReader {
+		return func(ctx context.Context, _ logrus.FieldLogger, _ *configpb.TestGroup, _ []InflatedColumn, _ time.Time, receivers chan<- InflatedColumn) error {
+			ctx, cancel := context.WithCancel(ctx)
+			defer cancel() // do not leak go routines
```
fejta, line 1066: This function does not start any go routines that can leak, so just using the ctx passed into the function is fine.

fejta (approved):

pkg/updater/updater_test.go
```diff
@@ -1059,6 +1059,40 @@ func TestInflateDropAppend(t *testing.T) {
...
+			for i := len(builds) - 1; i >= 0; i-- {
+				b := builds[i]
+				started := metadata.Started{}
```
fejta, line 1069: `var started metadata.Started`

PR #602, opened 2021-08-02
TestGrid Dynamic List - support extract email list from []interface{}

yairnee:
When reading list of strings from json go recognize it as a list of interfaces, which cause casting issues.
Added a function that deal with the case if it's a list of interfaces or list of strings. if not it's throw an error

### DIFF
--- a/metadata/job.go
+++ b/metadata/job.go
@@ -131,10 +131,10 @@ func (m Metadata) Strings() map[string]string {
 	return bm
 }
 
-/* GetListOfStrings get list of strings if exist, and true if they key is present.
+/* MultiString get list of strings if exist, and true if they key is present.
 If value is list of strings we return it as is.
 If value is list of interfaces, we try convert it into list of strings, if fail we return an empty list */
-func (m Metadata) GetListOfStrings(name string) ([]string, bool) {
+func (m Metadata) MultiString(name string) ([]string, bool) {
 	if v, ok := m[name]; !ok {
 		return []string{}, false
 	} else if lstStr, good := v.([]string); good {
--- a/metadata/job.go
+++ b/metadata/job.go
@@ -17,7 +17,6 @@ limitations under the License.
 package metadata
 
 import (
-	"fmt"
 	"strings"
 )
 
@@ -132,29 +131,24 @@ func (m Metadata) Strings() map[string]string {
 	return bm
 }
 
-// ConvertToListOfStrings convert an interface to list of strings, support if the type of the interface
-// is list of strings or list of interfaces, where each interface in the list is a string.
-func ConvertToListOfStrings(rawInterface interface{}) ([]string, error) {
-	switch t := rawInterface.(type) {
-	case []string:
-		return t, nil
-	case []interface{}:
-		emails := []string{}
-		for _, item := range t {
-			switch tt := item.(type) {
-			case string:
-				emails = append(emails, tt)
-				break
-			default:
-				return []string{}, fmt.Errorf("one of the items in the list of the"+
-					" interfaces is not a string. value: '%v' type: %T", tt, tt)
+// GetListOfStrings get list of strings if exist, and true if they key is present.
+func (m Metadata) GetListOfStrings(name string) ([]string, bool) {
+	if v, ok := m[name]; !ok {
+		return []string{}, false
+	} else if lstStr, good := v.([]string); good {
+		return lstStr, true
+	} else if lstInter, good := v.([]interface{}); good {
+		convertedStrings := []string{}
+		for _, inter := range lstInter {
+			if s, good := inter.(string); !good {
+				return []string{}, true
+			} else {
+				convertedStrings = append(convertedStrings, s)
 			}
 		}
-		return emails, nil
-	default:
-		return []string{}, fmt.Errorf("rawInterface type is not list of "+
-			"strings or interfaces. value: '%v' type: %T", t, t)
+		return convertedStrings, true
 	}
+	return []string{}, true
 }
 
 // firstFilled returns the first non-empty option or else def.
--- a/metadata/job.go
+++ b/metadata/job.go
@@ -131,7 +131,9 @@ func (m Metadata) Strings() map[string]string {
 	return bm
 }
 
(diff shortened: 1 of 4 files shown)


pkg/updater/gcs.go
```diff
@@ -390,19 +390,44 @@ func convertResult(log logrus.FieldLogger, nameCfg nameConfig, id string, header
 
 	emailAddressesInterface, ok := result.finished.Finished.Metadata[EmailListKey]
 	if ok {
-		emailAddresses, ok := emailAddressesInterface.([]string)
-		if ok {
-			out.Column.EmailAddresses = emailAddresses
-		} else {
-			log.Error("email addresses type is not list of strings")
+		emailAddresses, err := convertToListOfStrings(emailAddressesInterface)
```
michelle192837, line 393: I think you can simplify this by just declaring the []string variable and using json.Marshal to coerce the value into that. e.g.

var []string emailAddresses
if err := json.Unmarshal(emailAddressesInterface, &emailAddresses); err != nil {
  log.WithError(err).Error("<error message>")
}

yairnee: Updated

fejta:
Seems like this PR should involve updating https://github.com/GoogleCloudPlatform/testgrid/blob/e5fd38ec0774b287c20c1147d90f9cd064739f48/metadata/job.go#L85-L132

pkg/updater/gcs.go
```diff
@@ -390,12 +391,17 @@ func convertResult(log logrus.FieldLogger, nameCfg nameConfig, id string, header
...
 	emailAddressesInterface, ok := result.finished.Finished.Metadata[EmailListKey]
 	if ok {
-		emailAddresses, ok := emailAddressesInterface.([]string)
-		if ok {
-			out.Column.EmailAddresses = emailAddresses
-		} else {
-			log.Error("email addresses type is not list of strings")
+		// we do marshal and unmarshal to json to convert the result from []interface{} to []string
+		var emailAddresses []string
+		emailAddressesData, err := json.Marshal(emailAddressesInterface)
```
fejta, line 396: This is a horribly inefficient way to do this.

pkg/updater/gcs.go
```diff
@@ -390,12 +391,17 @@ func convertResult(log logrus.FieldLogger, nameCfg nameConfig, id string, header
 
 	emailAddressesInterface, ok := result.finished.Finished.Metadata[EmailListKey]
 	if ok {
-		emailAddresses, ok := emailAddressesInterface.([]string)
```
fejta, line 393: Do we have unit tests that cover this scenario? Can we add them?

pkg/updater/gcs.go
```diff
@@ -390,12 +391,17 @@ func convertResult(log logrus.FieldLogger, nameCfg nameConfig, id string, header
 
 	emailAddressesInterface, ok := result.finished.Finished.Metadata[EmailListKey]
```
fejta, line 392: There are strings inside the `[]interface{}`, so you can do this:

```golang
var items []interface{}
for _, item := range items {
  s, ok := item.(string)
}
```

https://play.golang.org/p/vLuYFf_mUpB

yairnee:
> Seems like this PR should involve updating
> 
> https://github.com/GoogleCloudPlatform/testgrid/blob/e5fd38ec0774b287c20c1147d90f9cd064739f48/metadata/job.go#L85-L132

update how? the logic of the EmailAddress into the job file?  It's not too specific?

pkg/updater/gcs.go
yairnee: The email list is supposed to be amount of committers, not should be  a lot of  them.
But I had another method here that is more efficient but also more complex

https://github.com/GoogleCloudPlatform/testgrid/pull/602/files/cde6941ddad425290e448cfe83672ff3494f5468

pkg/updater/gcs.go
yairnee: Not sure I understand, It's what I removed

pkg/updater/gcs.go
fejta: This should call one of the Metadata methods, rather than accessing the map.

fejta:
> update how? the logic of the EmailAddress into the job file? It's not too specific?

There's nothing specific about email addresses to this change. It's about returning a list of strings rather than a single string or a map of strings.

yairnee:
> > update how? the logic of the EmailAddress into the job file? It's not too specific?
> 
> There's nothing specific about email addresses to this change. It's about returning a list of strings rather than a single string or a map of strings.

Oh cool, this is why I made that function generic, updated

metadata/job.go
```diff
@@ -131,6 +132,31 @@ func (m Metadata) Strings() map[string]string {
 	return bm
 }
 
+// ConvertToListOfStrings convert an interface to list of strings, support if the type of the interface
+// is list of strings or list of interfaces, where each interface in the list is a string.
+func ConvertToListOfStrings(rawInterface interface{}) ([]string, error) {
```
yairnee, line 137: TODO: add UT

fejta: Please make this a member of the Metadata struct, maybe MultiString instead of String?

yairnee: Made as Metadata struct function, also kept the logic on the other functions there, which return value and found for key.
It's might cause issues, because the user will not be able to distinguish between an empty list, or wasn't able to convert the list of interfaces into list of strings. 
WDYT? 

I added UT's.
Not sure about the MultiString you mean for the DocString? you mean maybe Multilines?

metadata/job.go
```diff
@@ -131,6 +131,28 @@ func (m Metadata) Strings() map[string]string {
 	return bm
 }
 
+/* GetListOfStrings get list of strings if exist, and true if they key is present.
+If value is list of strings we return it as is.
+If value is list of interfaces, we try convert it into list of strings, if fail we return an empty list */
+func (m Metadata) GetListOfStrings(name string) ([]string, bool) {
```
fejta, line 137: Can you pick a name that more closely follows the other methods? Otherwise LGTM!

fejta (commented):
/lint

fejta (approved):

PR #626, opened 2021-08-11
Add implementation for ListHeaders, ListRows

abdelrady:
Add implementation for ListHeaders, ListRows

* /dashboards/{dashboard}/tabs/{tab}/headers -> ListHeaders
* /dashboards/{dashboard}/tabs/{tab}/rows -> ListRows

### DIFF
--- a/pkg/api/v1/config_test.go
+++ b/pkg/api/v1/config_test.go
@@ -25,11 +25,13 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"testing"
+	"time"
 
 	"cloud.google.com/go/storage"
 	"github.com/golang/protobuf/proto"
 
 	pb "github.com/GoogleCloudPlatform/testgrid/pb/config"
+	statepb "github.com/GoogleCloudPlatform/testgrid/pb/state"
 	"github.com/GoogleCloudPlatform/testgrid/util/gcs"
 )
 
@@ -123,9 +125,10 @@ func TestPassQueryParameters(t *testing.T) {
 	}
 }
 
-type ConfigTestSpec struct {
+type TestSpec struct {
 	name             string
 	config           map[string]*pb.Configuration
+	grid             map[string]*statepb.Grid
 	endpoint         string
 	params           string
 	expectedResponse string
@@ -133,7 +136,7 @@ type ConfigTestSpec struct {
 }
 
 func TestListDashboardGroups(t *testing.T) {
-	tests := []ConfigTestSpec{
+	tests := []TestSpec{
 		{
 			name: "Returns an empty JSON when there's no groups",
 			config: map[string]*pb.Configuration{
@@ -200,7 +203,7 @@ func TestListDashboardGroups(t *testing.T) {
 }
 
 func TestGetDashboardGroup(t *testing.T) {
-	tests := []ConfigTestSpec{
+	tests := []TestSpec{
 		{
 			name: "Returns an error when there's no resource",
 			config: map[string]*pb.Configuration{
@@ -300,7 +303,7 @@ func TestGetDashboardGroup(t *testing.T) {
 }
 
 func TestListDashboards(t *testing.T) {
-	tests := []ConfigTestSpec{
+	tests := []TestSpec{
 		{
 			name: "Returns an empty JSON when there is no dashboards",
 			config: map[string]*pb.Configuration{
@@ -367,7 +370,7 @@ func TestListDashboards(t *testing.T) {
 }
 
 func TestGetDashboard(t *testing.T) {
-	tests := []ConfigTestSpec{
+	tests := []TestSpec{
 		{
 			name: "Returns an error when there's no resource",
 			config: map[string]*pb.Configuration{
@@ -500,7 +503,7 @@ func TestGetDashboard(t *testing.T) {
 }
 
 func TestGetDashboardTabs(t *testing.T) {
-	tests := []ConfigTestSpec{
+	tests := []TestSpec{
 		{
 			name: "Returns an error when there's no resource",
 			config: map[string]*pb.Configuration{
@@ -626,10 +629,10 @@ func TestGetDashboardTabs(t *testing.T) {
 	RunTestsAgainstEndpoint(t, "/dashboards/", tests)
 }
 
-func RunTestsAgainstEndpoint(t *testing.T, baseEndpoint string, tests []ConfigTestSpec) {
+func RunTestsAgainstEndpoint(t *testing.T, baseEndpoint string, tests []TestSpec) {
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
-			router := Route(nil, setupTestServer(t, test.config))
+			router := Route(nil, setupTestServer(t, test.config, test.grid))
 			request, err := http.NewRequest("GET", baseEndpoint+test.endpoint+test.params, nil)
 			if err != nil {
 				t.Fatalf("Can't form request: %v", err)
@@ -650,7 +653,7 @@ func RunTestsAgainstEndpoint(t *testing.T, baseEndpoint string, tests []ConfigTe
 // Helper Functions
 ///////////////////
 
-func setupTestServer(t *testing.T, configurations map[string]*pb.Configuration) Server {
+func setupTestServer(t *testing.T, configurations map[string]*pb.Configuration, grids map[string]*statepb.Grid) Server {
 	t.Helper()
 
 	var fc fakeClient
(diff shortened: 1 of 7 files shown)


abdelrady:
@chases2 I appreciate your review to this draft PR. I want to make sure I'm on the right track on serving tabs headers/rows data.

pkg/api/v1/state.go
```diff
@@ -0,0 +1,232 @@
...
+	if err != nil {
+		return nil, time.Time{}, 0, fmt.Errorf("client.Open(): %w", err)
+	}
+	if attrs == nil {
+		return r, time.Time{}, 0, nil
+	}
+	return r, attrs.LastModified, attrs.Generation, nil
+}
+
+// getDashboardTabInternal locate dashboard tab in config
```
michelle192837, line 84: nit: Don't think 'internal' is needed. 'findDashboardTab' maybe?

also nit: 'locates'?

pkg/api/v1/state.go
```diff
@@ -0,0 +1,232 @@
...
+	cfg := s.getConfig(w, r)
+	if cfg == nil {
+		return
+	}
+
+	grid, errMsg := s.getDashboardTabGrid(r, cfg, vars["dashboard"], vars["tab"])
+	if errMsg != "" {
+		http.Error(w, errMsg, http.StatusNotFound)
+	}
+	if grid != nil && grid.Columns != nil {
```
michelle192837, line 171: Prefer early error returns (https://danp.net/posts/reducing-go-nesting/); check if nil and return if so.

Also, not sure if grid.Columns = nil is an error; think we can just return an empty response instead if the columns or grid are empty, instead of not found.

pkg/api/v1/state.go
```diff
@@ -0,0 +1,232 @@
...
+// GetTabHeaders returns dashboard tab headers
+// Response json: ListHeadersResponse
+func (s Server) GetTabHeaders(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	cfg := s.getConfig(w, r)
+	if cfg == nil {
+		return
+	}
+
+	grid, errMsg := s.getDashboardTabGrid(r, cfg, vars["dashboard"], vars["tab"])
```
michelle192837, line 167: errMsg can just be `err` (here and elsewhere)

pkg/api/v1/state.go
```diff
@@ -0,0 +1,232 @@
...
+		var dashboardTabResponse apipb.ListRowsResponse
+		for _, gRow := range grid.Rows {
+			row := apipb.ListRowsResponse_Row{
+				Name:   gRow.Name,
+				Issues: gRow.Issues,
+				Alert:  gRow.AlertInfo,
+			}
+			cellsCount := len(gRow.CellIds)
+			gRowDecodedResults := decodeRLE(gRow.Results)
+			// loop throug CellIds, Messages, Icons slices and build cell struct objects
```
michelle192837, line 216: nit:"through"

pkg/api/v1/state.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+	configPath, _, _ := s.configPath(r)
+	grid, err := s.getTabGrid(configPath, cfg, tab.TestGroupName)
+	if err != nil {
+		return nil, fmt.Sprintf("%v", err)
+	}
+	return grid, ""
+}
+
+// decodeRLE decode the run length encoded data
```
michelle192837, line 143: nit: "decodes"

pkg/api/v1/state.go
```diff
@@ -0,0 +1,232 @@
...
+//   [0, 3, 5, 4] -> [0, 0, 0, 5, 5, 5, 5]
+func decodeRLE(encodedData []int32) []int32 {
```
michelle192837, line 145: We kiiiinda have this in https://github.com/GoogleCloudPlatform/testgrid/blob/c1f9367972d6fccb9a1d94112560acff469d3379/internal/result/results.go#L95, but not quite. Probably fine to keep this method as-is.

michelle192837 (commented):
Thanks! Have some general comments, and I think state.go probably warrants unit testing as well, but broadly looks good.

pkg/api/v1/server.go
```diff
@@ -40,5 +41,8 @@ func Route(r *mux.Router, s Server) *mux.Router {
 	r.HandleFunc("/dashboards", s.ListDashboards).Methods("GET")
 	r.HandleFunc("/dashboards/{dashboard}/tabs", s.ListDashboardTabs).Methods("GET")
 	r.HandleFunc("/dashboards/{dashboard}", s.GetDashboard).Methods("GET")
+
+	r.HandleFunc("/dashboards/{dashboard}/tabs/{tab}/headers", s.GetTabHeaders).Methods("GET")
```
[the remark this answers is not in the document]
abdelrady: Fixed.

pkg/api/v1/state.go
abdelrady: Yes, Unfortunately the existing method doesn't provide required return type ([]int32) by the proto. So I will keep it for now.

pkg/api/v1/state.go
```diff
@@ -0,0 +1,232 @@
...
+				}
+			}
+			return nil
+		}
+	}
+	return nil
+}
+
+// getTabGrid fetch tab group name grid info (columns, rows, ..etc)
+func (s Server) getTabGrid(configPath *gcs.Path, cfg *configpb.Configuration, groupName string) (*statepb.Grid, error) {
```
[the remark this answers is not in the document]
abdelrady: Good to know. Thanks for sharing! Done.

pkg/api/v1/state.go
abdelrady: That makes sense. Updated.

abdelrady:
@michelle192837 @chases2 This PR is ready for another round of review. Thank you!

pkg/api/v1/state.go
```diff
@@ -0,0 +1,174 @@
...
+	return grid, err
+}
+
+// Grid fetch tab and grid info (columns, rows, ..etc)
+func (s Server) Grid(r *http.Request, cfg *configpb.Configuration, dashboardKey string, tabKey string) (*statepb.Grid, string) {
+	tab := findDashboardTab(cfg, dashboardKey, tabKey)
+	if tab == nil {
+		return nil, fmt.Sprintf("Dashboard {%q} or tab {%q} not found", dashboardKey, tabKey)
+	}
+	configPath, _, _ := s.configPath(r)
```
chases2, line 73: If this generated an error, seeing why early would be useful.

pkg/api/v1/state.go
```diff
@@ -0,0 +1,174 @@
...
+		return
+	}
+
+	grid, err := s.Grid(r, cfg, vars["dashboard"], vars["tab"])
+	if err != "" {
+		http.Error(w, err, http.StatusNotFound)
+		return
+	}
+	if grid == nil {
+		http.Error(w, fmt.Sprintf("grid not found."), http.StatusNotFound)
```
chases2, line 111: nit: unnecessary fmt.Sprintf call

chases2 (commented):
Overall, this is great! I'd like to see more logging or info messages in-general to make debugging easier.

pkg/api/v1/config_test.go
```diff
@@ -126,6 +128,7 @@ func TestPassQueryParameters(t *testing.T) {
 type ConfigTestSpec struct {
```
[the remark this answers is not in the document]
abdelrady: I agree. Done.

pkg/api/v1/state.go
```diff
@@ -0,0 +1,174 @@
...
+// GetTabHeaders returns dashboard tab headers
+// Response json: ListHeadersResponse
+func (s Server) ListHeaders(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	cfg := s.getConfig(w, r)
+	if cfg == nil {
+		return
+	}
+
+	grid, err := s.Grid(r, cfg, vars["dashboard"], vars["tab"])
```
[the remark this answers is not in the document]
abdelrady: That makes sense. Updated.

pkg/api/v1/state.go
abdelrady: I skipped this at first because cfg would be nil if s.configPath() would generate an error. And there was a cfg nil check before that call. Now, I agree that if that cfg nil check is removed in the future, this would leave this call exposed to potential exceptions.
Updated.

pkg/api/v1/state.go
abdelrady: Good catch! Removed.

michelle192837 (commented):
Aside from lint errors, looks good to me! I'll leave final approval to Sean.

abdelrady:
@chases2 Comments resolved. Please let me know if you have any other concerns or comments.

chases2 (approved):

PR #629, opened 2021-08-11
Refactor pubsub package for easier unit testing.

fejta:
Create a wrapped client type, which implements a Susbscriber interface that will Subscribe and returns a Sender.

### DIFF
--- a/pkg/pubsub/pubsub.go
+++ b/pkg/pubsub/pubsub.go
@@ -28,6 +28,31 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+// Subscriber creates Senders that attach to subscriptions with the specified settings.
+type Subscriber interface {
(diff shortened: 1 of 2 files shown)


pkg/pubsub/pubsub.go
```diff
@@ -28,6 +28,31 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+// Subscriber creates Senders that attach to subscriptions with the specified settings.
+type Subscriber interface {
+	Subscribe(projID, subID string, setttings *pubsub.ReceiveSettings) Sender
+}
+
+// Client makes a pubsub client a Subscriber that can Send pubsub messages.
```
michelle192837, line 36: nit: feel like this description is missing a word?

michelle192837 (approved):

PR #635, opened 2021-08-12
PubSub bootstrapping scripts

fejta:


cluster/list-gcs-prefixes.sh
```diff
@@ -0,0 +1,25 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+if [[ $# -lt 1 ]]; then
+    echo "Usage: $(basename "$0") <gs://path/to/config>" >&2
+    exit 1
+fi
+go run github.com/GoogleCloudPlatform/testgrid/config/print --path="$1" \
```
michelle192837, line 21: nit: I assume this is printing config and grabbing all gcs_prefix entries, but a short comment would be nice (with an example raw line this is parsing?)

cluster/create-subscription.sh
```diff
@@ -0,0 +1,89 @@
...
+        c) canary=1;;
+    esac
+done
+
+shift $((OPTIND -1))
+
+if [[ $# -lt 1 || -z "$topic" ]]; then
+    echo "Usage: $(basename "$0") [-a $ack] [-e $exp] -c [-p $project] <-t TOPIC> SUBSCRIPTION [EXTRA ...]" >&2
+    echo >&2
+    echo "  -a ACK  : seconds to allow testgrid to ack a message" >&2
```
michelle192837, line 41: Should we recommend defaults for any of these? (I think everything but -p and maybe -t could have a reasonable default)

michelle192837 (approved):
/hold

cluster/create-subscription.sh
fejta: If you check  L22-23 and L39 you'll see that we actually do have a default and print it.

cluster/list-gcs-prefixes.sh
fejta: Done

fejta (commented):
/hold cancel

PR #646, opened 2021-08-12
Add --subscribe flags to prod.

fejta:
Also sort flags and update resource config to better match what autopilot is adjusting it to

/cc @chaodaiG  @cjwagner

### DIFF
--- a/cluster/prod/updater.yaml
+++ b/cluster/prod/updater.yaml
@@ -32,19 +32,22 @@ spec:
       - name: updater
         image: gcr.io/k8s-testgrid/updater:v20210812-v0.0.90-38-g463fa11
         args:
-        - --build-timeout=2m
         - --build-concurrency=2
-        - --group-concurrency=4
+        - --build-timeout=2m
         - --config=gs://k8s-testgrid/config
         - --confirm
-        - --json-logs
+        - --group-concurrency=4
         - --group-timeout=10m
-        - --wait=10m
+        - --json-logs
+        - --subscribe=kubernetes-jenkins=kubernetes-jenkins/testgrid
+        - --subscribe=oss-prow=k8s-testgrid/testgrid
+        - --subscribe=istio-prow=k8s-testgrid/testgrid
+        - --wait=2h
         resources:
           requests:
-            cpu: "1"
-            memory: "50G"
+            cpu: "5"
+            memory: "20G"
           limits:
-            cpu: "2"
-            memory: "100G"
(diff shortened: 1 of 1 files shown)


cluster/prod/updater.yaml
```diff
@@ -32,19 +32,22 @@ spec:
...
         - --confirm
-        - --json-logs
+        - --group-concurrency=4
         - --group-timeout=10m
-        - --wait=10m
+        - --json-logs
+        - --subscribe=kubernetes-jenkins=kubernetes-jenkins/testgrid
+        - --subscribe=oss-prow=k8s-testgrid/testgrid
+        - --subscribe=istio-prow=k8s-testgrid/testgrid
+        - --wait=2h
```
cjwagner, line 45: Could you please explain what the `subscribe` flags do and the motivation for the increased `wait` value? I'm not familiar with these flags.

fejta: It subscribes to notifications of changes in these buckets and immediately does the update.

fejta: And since we'll immediately detect when test groups have new results to process (and process them) we no longer need to waste so much time checking every 10m to see whether there are any new results.

cjwagner: Gotcha, what about the `wait` change from `10m` to `2h`? I see the help info for the flag here: https://github.com/GoogleCloudPlatform/testgrid/blob/d2aa8f45ec7eee6ece5376eeb8eb1199622be898/cmd/updater/main.go#L125
I'm curious why we'd want to increase the update period to only update once per 2 hours. Is that because this is a complete list of tenants and we're switching every tenant to subscription based updates (which ignore the wait)?

cjwagner (approved):
/hold

cluster/prod/updater.yaml
fejta: 1) Note that "at least."
  * Part of the purpose is to ensure we backoff if there's an outage -- waiting 10m to retry rather than instantly
2) We want to spend our compute time updating groups that have changed
  * Otherwise we have too many test groups to finish in 10m.
  * This is largely because we spend so much effort validating that tabs have not changed.
  * Instead pubsub allows us to detect which groups are changing and focus efforts on those groups
  * Continue to do periodic updates on occasion to protect from any odd pubsub behavior.

In order to improve performance we need to reduce the wait time. Otherwise the pubsub notifications can't help much because we're already hugely behind our processing.

PR #651, opened 2021-08-13
Match PR directory links.

fejta:
Presubmits tend to get uploaded in a `gs://prefix/directory/job/1234.txt` with a link to `gs://prefix/pull/123/job/1234`

### DIFF
--- a/pkg/updater/pubsub.go
+++ b/pkg/updater/pubsub.go
@@ -213,11 +213,15 @@ func processNotification(paths map[gcs.Path][]string, n *pubsub.Notification) ([
 	b, obj := n.Path.Bucket(), n.Path.Object()
 	base := path.Base(obj)
 	dur, ok := namedDurations[base]
-	if !ok { // Maybe it is a junit file, should finish soon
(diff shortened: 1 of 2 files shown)


pkg/updater/pubsub.go
```diff
@@ -214,10 +214,14 @@ func processNotification(paths map[gcs.Path][]string, n *pubsub.Notification) ([
 	base := path.Base(obj)
 	dur, ok := namedDurations[base]
 	if !ok { // Maybe it is a junit file, should finish soon
-		if !strings.HasPrefix(base, "junit") || !strings.HasSuffix(base, ".xml") {
+		switch {
+		case strings.HasPrefix(base, "junit") && strings.HasSuffix(base, ".xml"):
+			dur = 5 * time.Minute
+		case strings.HasSuffix(base, ".txt") && strings.Contains(obj, "directory/"):
```
michelle192837, line 220: nit: Consider a comment for what the new case covers (I keep forgetting these are symlinks?)

michelle192837 (approved):
/hold

pkg/updater/pubsub.go
fejta: Done

Issue #656, opened 2021-08-18
Successful prow job showed red `pod` on testgrid 


chaodaiG:
https://testgrid.k8s.io/googleoss-test-infra#autobump-prow-for-auto-deploy

![Screen Shot 2021-08-18 at 11 09 09 AM](https://user-images.githubusercontent.com/45011425/129950041-ec32831d-53c6-4741-9e2a-8f572c4d2cb3.png)

the prow job with red `pod` is https://oss-prow.knative.dev/view/gs/oss-prow/logs/ci-oss-test-infra-autobump-prow-for-auto-deploy/1428054460566867968

chaodaiG:
also happened at https://testgrid.k8s.io/sig-testing-prow#push-prow

fejta:
```
   77  2021-08-18T18:02:09Z  gs://oss-prow/logs/ci-oss-test-infra-autobump-prow-for-auto-deploy/1428054460566867968/finished.json#1629309729931792
24092  2021-08-18T18:02:28Z  gs://oss-prow/logs/ci-oss-test-infra-autobump-prow-for-auto-deploy/1428054460566867968/podinfo.json#1629309748149434
```


podinfo was uploaded 19s after finished.json

chaodaiG:
This should be fixed by https://github.com/GoogleCloudPlatform/testgrid/pull/657
/close

PR #657, opened 2021-08-18
Allow time to collect podinfo.

fejta:
The podinfo.json is updated by crier, some time after the sidecar uploads finished.json.
Detect this scenario and flag a missing podinfo.json as RUNNING until the result has reached its deadline:
* An hour after finished.json is uploaded or 24 hours after it started.

Also allow a larger buffer between when we receive a pubsub notification and when reprocess the tab.

/cc @chaodaiG 
ref #656

### DIFF
--- a/pkg/updater/pubsub.go
+++ b/pkg/updater/pubsub.go
@@ -195,10 +195,10 @@ func processGCSNotifications(ctx context.Context, log logrus.FieldLogger, q *con
 }
 
 var namedDurations = map[string]time.Duration{
-	"podinfo.json":  time.Second,     // Done
-	"finished.json": time.Minute,     // Container done, wait for prowjob to finish
-	"metadata.json": 5 * time.Minute, // Should finish soon
-	"started.json":  time.Second,     // Running
+	"podinfo.json":  30 * time.Second, // Done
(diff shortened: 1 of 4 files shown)


pkg/updater/pubsub.go
```diff
@@ -195,10 +195,10 @@ func processGCSNotifications(ctx context.Context, log logrus.FieldLogger, q *con
 }
 
 var namedDurations = map[string]time.Duration{
-	"podinfo.json":  time.Second,     // Done
-	"finished.json": time.Minute,     // Container done, wait for prowjob to finish
-	"metadata.json": 5 * time.Minute, // Should finish soon
-	"started.json":  time.Second,     // Running
+	"podinfo.json":  30 * time.Second, // Done
```
chaodaiG, line 198: why changing these values?

fejta: > Also allow a larger buffer between when we receive a pubsub notification and when reprocess the tab.

chaodaiG (approved):

pkg/updater/pubsub.go
fejta: Suspect we may get the pubsub notification before it appears in the `ls`

PR #662, opened 2021-08-21
Fix a panic in updater (if update errors and freq = 0)

michelle192837:
This generally only happens when running locally (otherwise freq/--wait is set to something reasonable), and only when there's an error in the update, but it is annoying when it happens.

### DIFF
--- a/pkg/updater/updater.go
+++ b/pkg/updater/updater.go
@@ -460,9 +460,14 @@ func Update(parent context.Context, client gcs.ConditionalClient, mets *Metrics,
 		}
 		unprocessed, err := update(ctx, client, log, tg, *tgp, updateGroup, write, gen, fin)
 		if err != nil {
-			delay := freq/4 + time.Duration(rand.Int63n(int64(freq/4)))
-			log.WithError(err).WithField("delay", delay).Error("Error updating group")
-			q.Fix(tg.Name, time.Now().Add(delay), true)
+			log := log.WithError(err)
+			var delay time.Duration
+			if freq > 0 {
+				delay = freq/4 + time.Duration(rand.Int63n(int64(freq/4))) // Int63n() panics if freq <= 0
+				log = log.WithField("delay", delay.Seconds())
+				q.Fix(tg.Name, time.Now().Add(delay), true)
+			}
+			log.Error("error updating group")
 			return
 		}
 		if unprocessed { // process another chunk ASAP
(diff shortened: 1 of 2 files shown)


pkg/updater/updater.go
```diff
@@ -460,7 +460,11 @@ func Update(parent context.Context, client gcs.ConditionalClient, mets *Metrics,
 		}
 		unprocessed, err := update(ctx, client, log, tg, *tgp, updateGroup, write, gen, fin)
 		if err != nil {
-			delay := freq/4 + time.Duration(rand.Int63n(int64(freq/4)))
+			// delay := freq/4 + time.Duration(rand.Int63n(int64(freq/4)))
```
chases2, line 463: Please delete this comment; the code isn't much use if it's panicky

chases2 (approved):
/lgtm
/hold for nits

pkg/updater/updater.go
```diff
@@ -460,7 +460,11 @@ func Update(parent context.Context, client gcs.ConditionalClient, mets *Metrics,
...
+			var delay time.Duration
+			if freq > 0 {
```
fejta, line 465: Please don't call Fix with a delay of zero -- meaning retry this error right away, flooding the system with these errors.

Something more like:

```golang
log := log.WithError(err) // shadow
if freq > 0 {
  dleay = blah
  log = log.WithField("delay", delay) // maybe delay.Seconds()?
  q.Fix(tg.Name, etc)
}
log.Error("error updating group)
```

pkg/updater/updater.go
```diff
@@ -460,9 +460,14 @@ func Update(parent context.Context, client gcs.ConditionalClient, mets *Metrics,
...
 		unprocessed, err := update(ctx, client, log, tg, *tgp, updateGroup, write, gen, fin)
 		if err != nil {
-			delay := freq/4 + time.Duration(rand.Int63n(int64(freq/4)))
-			log.WithError(err).WithField("delay", delay).Error("Error updating group")
-			q.Fix(tg.Name, time.Now().Add(delay), true)
+			log := log.WithError(err)
+			var delay time.Duration
+			if freq > 0 {
+				delay = freq/4 + time.Duration(rand.Int63n(int64(freq/4))) // Int63n() panics if freq <= 0
+				log.WithField("delay", delay.Seconds())
```
fejta, line 467: Currently this is a noop

log = log.WithField(...)

michelle192837:
Thanks folks!
/hold cancel

fejta (approved):
/hold

fejta (approved):

pkg/updater/updater.go
michelle192837: Whoops, done.

Issue #683, opened 2021-09-10
Column's "Started" should be a proto timestamp, not a double


chases2:
https://github.com/GoogleCloudPlatform/testgrid/blob/345d5369a1a438a11f7e28f979ef87247566b8e9/pb/state/state.proto#L119

aryans1319:
Can you assign this to me? I want to work for it

chases2:
Sure! You should be able to self-assign with the `/assign` command. See https://oss.gprow.dev/command-help for help with other sorts of commands, too.

/assign @aryans1319

nanu1605:
Hi @chases2, Is this issue up for grabs?

PR #684, opened 2021-09-10
Fix timestamp assignment from columns

chases2:
This float -> timestamp conversion is causing the displayed start times to be wrong when displayed, without (it seems) affecting the business logic of the updater. The rest of the updater seems to treat this field as milliseconds correctly.

It would be great if this proto field was simply a "timestamp", so it doesn't need to be converted to (and later from) a timestamp. (#683)


michelle192837:
I'd expect this to affect the alerter as well for new states, and that appears to be the case (looking at recent external-sent emails).

michelle192837 (approved):

PR #690, opened 2021-09-21
Truncate long messages when storing in state.

michelle192837:


pkg/updater/updater.go
```diff
@@ -1075,6 +1075,18 @@ func hasCellID(name string) bool {
 	return !strings.Contains(name, "@TESTGRID@")
 }
 
+// truncate truncates a message to max runes (and ellipses). Max = 0 returns the original message.
+func truncate(msg string, max int) string {
+	if max == 0 {
+		return msg
+	}
+	runes := []rune(msg)
```
fejta, line 1083: This seems unnecessary slow and ram hungry.

Don't we want the number of bytes, not runes? AKA this is about the storage size, not the number of characters.

Also, a) if len(s) < max then we already know len(runes) < max so there's no need to do the conversion

fejta: Also consider: https://pkg.go.dev/unicode/utf8#RuneCountInString

michelle192837: Since there's a lot of flex for state storage (people might/might not be adding a lot of metrics, messages, icons, etc.) I was thinking that cutting by runes is more readable, but it's prob. rare for just cutting bytes to not work for messages.

(Huh, and apparently 1.13+ has https://pkg.go.dev/strings#ToValidUTF8).

fejta (approved):
clever solution!

PR #692, opened 2021-09-23
Increase CPU, RAM and --build-concurrency

fejta:
ref https://github.com/kubernetes/test-infra/issues/23720


chaodaiG (approved):
/hold
please make sure that the node has at least 100G memory

fejta:
> please make sure that the node has at least 100G memory

This is an autopilot cluster, there are no nodes to manage.

PR #695, opened 2021-09-23
Allow InflateGrid to time out.

fejta:
This can be an expensive operation so check the context each column and abort early when it is done.

### DIFF
--- a/pkg/updater/inflate.go
+++ b/pkg/updater/inflate.go
@@ -73,11 +73,13 @@ type Cell struct {
 //
 // Drops columns before earliest or more recent than latest.
 // Also returns a map of issues associated with each row name.
-func InflateGrid(grid *statepb.Grid, earliest, latest time.Time) ([]InflatedColumn, map[string][]string) {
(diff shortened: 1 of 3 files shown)


pkg/updater/inflate.go
```diff
@@ -73,11 +73,14 @@ type Cell struct {
...
 // Drops columns before earliest or more recent than latest.
 // Also returns a map of issues associated with each row name.
-func InflateGrid(grid *statepb.Grid, earliest, latest time.Time) ([]InflatedColumn, map[string][]string) {
+func InflateGrid(ctx context.Context, grid *statepb.Grid, earliest, latest time.Time) ([]InflatedColumn, map[string][]string, error) {
 	var cols []InflatedColumn
+	if n := len(grid.Columns); n > 0 {
+		cols = make([]InflatedColumn, 0, n)
+	}
 
 	// nothing is blocking, so no need for a parent context.
```
michelle192837, line 82: Is this comment still true?

michelle192837 (approved):
/hold

pkg/updater/inflate.go
fejta: Done

PR #698, opened 2021-09-25
Populate FailBuildID based on data source

chases2:
Also adds a TODO I spotted in the summarizer for where/when these alerts should be created


pkg/updater/updater.go
```diff
@@ -1002,7 +1002,7 @@ func ConstructGrid(log logrus.FieldLogger, group *configpb.TestGroup, cols []Inf
 		})
 	}
 
-	alertRows(grid.Columns, grid.Rows, failsOpen, passesClose)
+	alertRows(grid.Columns, grid.Rows, failsOpen, passesClose, group.UseKubernetesClient)
```
fejta, line 1005: Please also check `result_source` here, which is another valid way of specifying that we should use this client.

https://github.com/GoogleCloudPlatform/testgrid/blob/5280086a466030e3a2c314364e8cbe6b9108eab6/pkg/updater/updater.go#L480-L488

fejta (approved):
/hold

PR #707, opened 2021-09-29
Add BenchmarkInflateGrid

fejta:
Run with: `go test ./pkg/updater -bench=.`

https://pkg.go.dev/testing#hdr-Benchmarks

### DIFF
--- a/pkg/updater/BUILD.bazel
+++ b/pkg/updater/BUILD.bazel
@@ -39,6 +39,7 @@ go_test(
         "read_test.go",
         "updater_test.go",
     ],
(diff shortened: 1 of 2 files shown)


michelle192837 (approved):
/hold

Is `pkg/updater/testdata/pull-kubernetes-unit.pb.z` supposed to be uploaded?

michelle192837 (approved):

fejta:
> Is pkg/updater/testdata/pull-kubernetes-unit.pb.z supposed to be uploaded?

Yes but I removed it

PR #722, opened 2021-10-07
Add pubsub integration to summarizer

fejta:
/hold

### DIFF
--- a/cmd/summarizer/main.go
+++ b/cmd/summarizer/main.go
@@ -21,14 +21,16 @@ import (
 	"errors"
 	"flag"
 	"runtime"
+	"strings"
 	"time"
 
-	"github.com/sirupsen/logrus"
-
+	gpubsub "cloud.google.com/go/pubsub"
+	"github.com/GoogleCloudPlatform/testgrid/pkg/pubsub"
+	"github.com/GoogleCloudPlatform/testgrid/pkg/summarizer"
 	"github.com/GoogleCloudPlatform/testgrid/util/gcs"
 	"github.com/GoogleCloudPlatform/testgrid/util/metrics"
-
-	"github.com/GoogleCloudPlatform/testgrid/pkg/summarizer"
+	"github.com/sirupsen/logrus"
+	"google.golang.org/api/option"
 )
 
 type options struct {
@@ -40,6 +42,7 @@ type options struct {
 	wait              time.Duration
 	gridPathPrefix    string
 	summaryPathPrefix string
+	pubsub            string
 
 	debug    bool
 	trace    bool
@@ -66,6 +69,7 @@ func gatherOptions() options {
 	flag.DurationVar(&o.wait, "wait", 0, "Ensure at least this much time has passed since the last loop (exit if zero).")
 	flag.StringVar(&o.gridPathPrefix, "grid-path", "grid", "Read grid states under this GCS path.")
 	flag.StringVar(&o.summaryPathPrefix, "summary-path", "summary", "Write summaries under this GCS path.")
+	flag.StringVar(&o.pubsub, "pubsub", "", "listen for test group updates at project/subscription")
 
 	flag.BoolVar(&o.debug, "debug", false, "Log debug lines if set")
 	flag.BoolVar(&o.trace, "trace", false, "Log trace and debug lines if set")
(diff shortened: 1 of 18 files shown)


cmd/summarizer/main.go
```diff
@@ -75,6 +79,23 @@ func gatherOptions() options {
 	return o
 }
 
+func gcsFixer(ctx context.Context, projectSub string, configPath gcs.Path, gridPrefix, credPath string) (summarizer.Fixer, error) {
+	if projectSub == "" {
+		return nil, nil
+	}
+	parts := strings.SplitN(projectSub, "/", 2)
+	if len(parts) != 2 {
+		return nil, errors.New("malformed project/subscription")
```
michelle192837, line 88: nit: Are the parameters this fails on noted anywhere? (I didn't see them here or when the error is caught in a parent function)

pkg/summarizer/summary.go
```diff
@@ -225,27 +211,128 @@ func lockDashboard(ctx context.Context, client gcs.ConditionalClient, path gcs.P
...
-		d := pathedDashboards[p]
-		dashboards[i] = d
-		generations[d.Name] = generationPaths[p]
+	for i := 0; i < concurrency; i++ {
+		go func() {
+			for dashName := range dashboardNames {
+				log := log.WithField("dashboard", dashName)
+				if err := updateName(log, dashName); err != nil {
+					mets.Error()
+					log.WithError(err).Error("Failed to summarize dasboard")
```
michelle192837, line 327: nit:dashboard

pkg/summarizer/summary.go
```diff
@@ -73,142 +70,131 @@ type gridReader func(ctx context.Context) (io.ReadCloser, time.Time, int64, erro
...
 
-	groupFinder := func(name string) (*configpb.TestGroup, gridReader, error) {
-		group := config.FindTestGroup(name, cfg)
-		if group == nil {
-			return nil, nil, nil
-		}
-		groupPath, err := configPath.ResolveReference(&url.URL{Path: path.Join(gridPathPrefix, name)})
-		if err != nil {
-			return group, nil, err
+	if dashboard != "" {
```
chases2, line 84: There's an opportunity in this loop to cut out all test groups that aren't the one this dashboard displays.
If it's worth trimming the config to one dashboard, it's probably worth trimming the test groups too.

pkg/summarizer/summary.go
```diff
@@ -73,142 +70,131 @@ type gridReader func(ctx context.Context) (io.ReadCloser, time.Time, int64, erro
...
+		name := dashboards[i].Name
+		switch {
+		case stat.Attrs != nil:
+			whens[name] = stat.Attrs.Updated.Add(freq)
+		default:
+			if errors.Is(stat.Err, storage.ErrObjectNotExist) {
+				wg.Add(1)
+				defer func(i int) {
+					defer wg.Done()
+					if _, err := lockDashboard(ctx, client, paths[i], 0); err != nil && !gcs.IsPreconditionFailed(err) {
```
chases2, line 167: Should we log here when err != nil? Is there something magic here? This seems identical to `lockDashboard(ctx, client, paths[i], 0)`

cmd/summarizer/main.go
fejta: Done

pkg/summarizer/pubsub.go
```diff
@@ -0,0 +1,113 @@
...
+			}).Trace("Fixing groups from gcs notifcation")
+			if err := q.FixTestGroups(when, false, *group); err != nil {
+				return err
+			}
+		}
+	}
+}
+
+// Return the test group associated with the notification.
+func processNotification(gridPrefix gcs.Path, n *pubsub.Notification) (*string, time.Duration) {
```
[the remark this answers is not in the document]
fejta: Done

pkg/summarizer/summary.go
fejta: We need to trim the dashboards otherwise we don't update them. 

The presence of extra testgroups is not a big deal right now. Since the logic to remove them is not straight forward so this seems cheaper and less error prone.

pkg/summarizer/summary.go
fejta: Done

pkg/summarizer/summary.go
fejta: Done

util/gcs/gcs.go
```diff
@@ -30,16 +30,29 @@ import (
...
 
 	statepb "github.com/GoogleCloudPlatform/testgrid/pb/state"
 
 	"cloud.google.com/go/storage"
 	"github.com/golang/protobuf/proto"
+	"google.golang.org/api/googleapi"
 	"google.golang.org/api/option"
 )
 
+func IsPreconditionFailed(err error) bool {
```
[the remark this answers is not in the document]
fejta: Done

fejta (commented):
/lint

PR #742, opened 2021-11-14
Convert config_merger logging to use Prometheus

chases2:
Adds prometheus tooling to our metrics utility and moves config_merger to use them

xref #744


chases2:
Should be ready for review; PTAL!

/assign @michelle192837 @fejta

michelle192837 (approved):
/hold

Unhold at will (or wait for Erick to review), but broadly looks good to me ^^

util/metrics/prometheus/prometheus.go
```diff
@@ -0,0 +1,139 @@
...
+	Values() map[string]float64
+}
+
+type int64Metric struct {
+	name   string
+	fields map[string]bool
+	met    *prometheus.GaugeVec
+}
+
+// init sets up the Prometheus endpoint at this port and URL when the package is imported
```
michelle192837, line 41: Should this 1) be an explicit initialization in the controller's main.go instead, and 2) be passed a port as an arg?

chases2: I think that hinges on "Do you think there should be a new `--metrics-port` flag on all controllers?" I lean towards "don't bother (yet)", but not very strongly.

If the port can be changed between binaries, we'd want a way to change it via a flag that should be in each main.go program. Otherwise, we should enforce that it's on the same port for everyone, and I think this is the best pattern for that.

michelle192837: I think "don't bother yet" seems fair, easy to add later!

fejta (approved):

PR #753, opened 2021-11-17
Add concurrence-safety tests to prometheus metrics

chases2:
Confirmed locally that race-detection tests pass with https://github.com/GoogleCloudPlatform/testgrid/pull/750 and fail without it.

/cc @fejta

### DIFF
--- a/util/metrics/prometheus/prometheus_test.go
+++ b/util/metrics/prometheus/prometheus_test.go
@@ -18,6 +18,7 @@ package prometheus
 
 import (
 	"strings"
+	"sync"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
@@ -55,7 +56,7 @@ func TestInt64Set(t *testing.T) {
 			},
 		},
 		{
-			name:   "values",
+			name:   "serial values",
 			fields: []string{"component"},
 			sets: []map[int64][]string{
 				{64: {"updater"}},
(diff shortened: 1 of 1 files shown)


util/metrics/prometheus/prometheus_test.go
```diff
@@ -96,10 +92,16 @@ func TestInt64Set(t *testing.T) {
...
-					m.Set(n, fields...)
+					wg.Add(1)
+					go func(n int64, fields []string) {
+						m.Set(n, fields...)
+						wg.Done()
+					}(n, fields)
 				}
+				wg.Wait()
 			}
 			got := m.(Valuer).Values()
```
fejta, line 106: Ideally we also ensure concurrently accessing both `Set()` and `Values()` works (aka concurrently reading from a map while someone else is writing to it)

See https://go.dev/blog/maps#concurrency

fejta (approved):
/hold

Thanks for these tests!

fejta (approved):

util/metrics/prometheus/prometheus_test.go
chases2: I considered writing a separate test for this, but I found that what I wanted to test was surprisingly small.

Ex. If I set a metric to "1", and then set it to "2" and read in parallel, I don't think I care which one the read returns; I just care that it doesn't panic, and that the final read is correct (which I test already).

This seemed like the easiest way to do reads and writes in parallel so that -race can detect if a mutex is broken. Thoughts?

PR #761, opened 2021-11-18
Standardize cycle metrics across controllers

chases2:
Adds duration controller to summarizer
Adds success/failure metrics to config_merger
Moves cycle timer to util/
Creates factory to standardize metric creation in main.go

/lint


michelle192837 (approved):
Think this looks good, unhold at will!

PR #769, opened 2021-11-23
Create Duration metric

chases2:
A common TestGrid metric is a duration, which is currently being stored as an Int64 number of seconds. This rendered anything that happens in less than one second (ex: most Config Mergers) invisible.

This change adds a metric type that measures elapsed time with the standard "duration" type.

### DIFF
--- a/util/metrics/prometheus/prometheus.go
+++ b/util/metrics/prometheus/prometheus.go
@@ -20,6 +20,7 @@ import (
 	"net/http"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/GoogleCloudPlatform/testgrid/util/metrics"
 	"github.com/prometheus/client_golang/prometheus"
@@ -27,13 +28,21 @@ import (
 	dto "github.com/prometheus/client_model/go"
 )
 
+func NewFactory() metrics.Factory {
+	return metrics.Factory{
+		NewInt64:    NewInt64,
+		NewCounter:  NewCounter,
+		NewDuration: NewDuration,
+	}
+}
+
 // Valuer extends a metric to include a report on its values.
 type Valuer interface {
 	metrics.Metric
 	Values() map[string]float64
 }
 
-type int64Metric struct {
+type gaugeMetric struct {
 	name   string
 	fields map[string]bool
 	met    *prometheus.GaugeVec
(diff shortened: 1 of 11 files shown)


util/metrics/prometheus/prometheus.go
```diff
@@ -148,3 +157,26 @@ func (m *counterMetric) Values() map[string]float64 {
 	}
 	return values
 }
+
+// Clock sets the metric's value to the given duration in seconds
+func (m *gaugeMetric) Clock(t time.Duration, fields ...string) {
+	m.lock.Lock()
+	defer m.lock.Unlock()
+	m.met.WithLabelValues(fields...).Set(t.Seconds())
```
fejta, line 165: Can we have `gaugeMetric` track the units desired and scale accordingly, rather than continuing to force everything into seconds?

fejta (commented):
/assign

util/metrics/prometheus/prometheus.go
chases2: Sure, I suppose. Prometheus doesn't seem to have a way to assign "metadata" in this way, other than changing the description.

I could change this interface so that NewDuration takes a unit (seconds, milliseconds), and then includes that unit with the description so we avoid a unitless number, but it didn't seem helpful to support `200 milliseconds` instead of `.2 seconds` for individual controllers.

fejta: I want to support 200 milliseconds

fejta: Given that we can spend anywhere from 10s to 20m updating a group, seconds seems like an appropriate unit (why I changed it from milliseconds).

OTOH when measuring latency that averages 50ms, reporting this as 50ms rather than 0.005ms tends to be easier to visualize and reason about.

fejta: Hmm, maybe that was just this https://github.com/GoogleCloudPlatform/testgrid/compare/fa39445790be66e9d611f32dced19728fd673ff9..fa39445790be66e9d611f32dced19728fd673ff9~1

fejta (approved):
/hold

I think having correct units for the problem being measured is useful and important, but do not need to block on this.

fejta (approved):

PR #772, opened 2021-12-01
Unescape local filepaths.

michelle192837:
Tested this locally, this is a quick and dirty fix for when a test group name contains a space (which is encoded by url.Parse).

### DIFF
--- a/util/gcs/local_gcs.go
+++ b/util/gcs/local_gcs.go
@@ -17,14 +17,17 @@ limitations under the License.
 package gcs
 
 import (
-	"cloud.google.com/go/storage"
 	"context"
-	"google.golang.org/api/iterator"
 	"io"
 	"io/ioutil"
+	"net/url"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
+
+	"cloud.google.com/go/storage"
+	"google.golang.org/api/iterator"
 )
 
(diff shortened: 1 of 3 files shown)


bweston92:
I might be talking nonsense here and I can't find any configuration to see how TestGrid is being configured, 
but if this is being configured to use no host-name (which I assume it is) then 1 or 3 slashes should be used. https://en.wikipedia.org/wiki/File_URI_scheme#How_many_slashes?

If they're using the correct file protocol this logic would do the following:

- `file:/abc/def/ghi.ext` => `file:/abc/def/ghi.ext`
- `file:///abc/def/ghi.ext` => `//abc/def/ghi.ext`

But I believe the expected to be:

- `file:/abc/def/ghi.ext` => `/abc/def/ghi.ext`
- `file:///abc/def/ghi.ext` => `/abc/def/ghi.ext`

util/gcs/local_gcs.go
```diff
@@ -45,7 +47,13 @@ func convertIsNotExistsErr(err error) error {
 }
 
 func cleanFilepath(path Path) string {
```
fejta, line 49: Unit tests?

fejta (approved):
/hold

fejta (approved):

PR #775, opened 2021-12-22
Add cache backing ReadGCS

chases2:
Modifies the Read GCS function to keep a cache of recently-requested configs. Mostly serves to speed up the API.

Controllers and commands using these functions:
`ReadGCS`: Config Merger, API
`Read`: Config Print, State Comparer

The Summarizer and Updater have their own logic to periodically recheck the configuration (not just as-needed); all of these could be collapsed into a config file subscription.

### DIFF
--- a/config/cache_test.go
+++ b/config/cache_test.go
@@ -1,0 +1,124 @@
+/*
+Copyright 2021 The TestGrid Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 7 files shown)


config/cache_test.go
```diff
@@ -0,0 +1,133 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
michelle192837, line 2: Update for this year and TestGrid Authors?

michelle192837: Same for the other file.

michelle192837 (approved):

config/cache_test.go
chases2: Done

michelle192837 (approved):

PR #777, opened 2022-01-10
Added simple autobug field for arbitrary extra text

KatrinaHoffert:
This is the schema change only. Will be implemented server-side separately.


KatrinaHoffert:
Hmm, test failure is `protos changed. Run bazel run //hack:update-protos`, but that gets

```
ERROR: An error occurred during the fetch of repository 'pip_deps':
   pip_import failed:  (src/main/tools/process-wrapper-legacy.cc:75: "execvp(python, ...)": No such file or directory
)
ERROR: no such package '@pip_deps//': pip_import failed:  (src/main/tools/process-wrapper-legacy.cc:75: "execvp(python, ...)": No such file or directory
)
```

KatrinaHoffert:
Seems that command requires python2 to be installed (ew, I did have python3 install).

michelle192837 (approved):

PR #778, opened 2022-01-10
Document update-protos requiring python2

KatrinaHoffert:
Not actually sure if it strictly requires Python 2 or what, but that's what fixed it for me.
There may be other requirements that aren't obvious to me (e.g., is the Go compiler
needed?).


michelle192837 (approved):
Thanks for catching!

PR #783, opened 2022-01-21
Allow multiple ccs in default test metadata

chases2:


fejta (commented):
Changing the type of the field 3 from a string to a list seems fraught.

How about putting the repeated field in slot 4?

AKA: 

```
message DefaultTestMetadata {
  string cc = 3;
  repeated string ccs = 4;
}
```

```golang
func getCCs() []string{
  return append(whatever.Ccs, whatever.Cc)
}
```

fejta (approved):
(the field is already repeated in some places we use it, so this makes this sort of problem less likely)

PR #789, opened 2022-01-29
Refactor API to support gRPC definitions

chases2:
Allows TestGrid API to be installed as a gRPC service as well as a web service.

Behavior change: Reports 404 in all cases, even when the config is unreadable or unauthorized.


fejta (approved):
/hold

This adds gRPC support but keeps the existing json responses?

PR #791, opened 2022-02-01
Minor go vet changes

chases2:
Test no longer copies RWMutex
Removes redundant nil check
/lint


fejta:
Please fix the error/flake

fejta (approved):

PR #802, opened 2022-02-05
Add persistent queue to summarizer.

fejta:
* Allow summarizer to accept multiple fixers
* Move util.Queue to util.queue.Queue
* Factor most FixPersistent logic into util.queue package
* Restart summary fixers when the configuration changes.
* Protect summarizer from a dashboard that no longer exists

### DIFF
--- a/pkg/summarizer/persist.go
+++ b/pkg/summarizer/persist.go
@@ -1,0 +1,19 @@
+package summarizer
+
+import (
+	"context"
+	"time"
+
+	"github.com/GoogleCloudPlatform/testgrid/config"
+	"github.com/GoogleCloudPlatform/testgrid/util/gcs"
+	"github.com/GoogleCloudPlatform/testgrid/util/queue"
+	"github.com/sirupsen/logrus"
+)
+
+// FixPersistent persists the updater queue using queue.FixPersistent.
+func FixPersistent(log logrus.FieldLogger, client queue.PersistClient, path gcs.Path, tick <-chan time.Time) Fixer {
+	fix := queue.FixPersistent(log, client, path, tick)
+	return func(ctx context.Context, iq *config.DashboardQueue) error {
+		return fix(ctx, &iq.Queue)
+	}
+}
--- a/pkg/summarizer/persist.go
+++ b/pkg/summarizer/persist.go
@@ -1,3 +1,19 @@
+/*
+Copyright 2022 The TestGrid Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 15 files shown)


pkg/summarizer/persist.go
```diff
@@ -1,3 +1,19 @@
+/*
+Copyright 2022 The Kubernetes Authors.
```
michelle192837, line 2: nit: TestGrid Authors (here and in the other file)

fejta: Done

michelle192837 (approved):

pkg/summarizer/summary.go
```diff
@@ -129,67 +129,99 @@ func Update(ctx context.Context, client gcs.ConditionalClient, mets *Metrics, co
...
 		}
 		return nil
 	}
+	cfgChanged := make(chan *snapshot.Config)
 
+	onConfigUpdate := func(cfg *snapshot.Config) error {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case cfgChanged <- cfg:
```
fejta, line 191: I think this observer pattern will work well @chases2!

chases2: A channel does seem more natural than a callback in golang; a lot fewer race conditions.

Feel free to merge this as-is, but I'm now looking to refactor Observe to use channels instead soon

chases2 (approved):

util/queue/persist.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+// Fixer will adjust the queue until the context expires.
+type Fixer func(context.Context, *Queue) error
+
+// FixPersistent persists the queue to the remote path every tick.
+//
+// The first time it will load the state. Thereafter it will save the state.
+// This includes restarts due to expiring contexts -- it will just load once.
+func FixPersistent(logr logrus.FieldLogger, client PersistClient, path gcs.Path, tick <-chan time.Time) Fixer {
```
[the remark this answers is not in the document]
fejta: (a) and (b)

I originally had this as `time.Duration` instead of `<-chan time.TIme` but then this made testing flaky and requiring sleeps. With this pattern we can test immediately and precisely:

```golang

fix := Fix(tc.client) // initial stored state
wg.Add(1)
go func() {
  for _, tick := range tc.ticks {
    ch <- tick
  }
  wg.Done()
}
fix(tc.q) // initial q state + fixing ticks
wg.Wait()
// verify uploaded files
// verify queue state
```


The signature isn't that much more complicated: `FixPersisent(..., time.NewTicker(time.Minute))` instead of `FixPersistent(..., time.Minute)` so seems like a good trade off for the increase in testing control.

PR #804, opened 2022-02-08
Refactor config snapshot to use channels

chases2:
/assign @fejta

### DIFF
--- a/config/snapshot/config_snapshot.go
+++ b/config/snapshot/config_snapshot.go
@@ -19,7 +19,6 @@ package snapshot
 import (
 	"context"
 	"fmt"
-	"sync"
 	"time"
 
 	"cloud.google.com/go/storage"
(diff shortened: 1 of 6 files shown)


config/snapshot/config_snapshot.go
```diff
@@ -29,130 +28,67 @@ import (
...
-			case <-ticker.C:
+			case <-ticker:
+				snap, err := updateHash(ctx, client, configPath, &cond)
+				switch {
+				case err == nil:
+					// Configuration changed
+					ch <- snap
+					cond.GenerationNotMatch = snap.Attrs.Generation
+				case !gcs.IsPreconditionFailed(err):
+					log.WithError(err).Errorf("Error while observing config")
```
fejta, line 73: Use `Error()` as you are not formatting anything.

config/snapshot/config_snapshot.go
```diff
@@ -29,130 +28,67 @@ import (
...
+					return
```
fejta, line 74: This doesn't seem robust enough -- a network flake, etc, will turn off this thread. I would expect it to try again next tick. IMO it should continue working until the context expires.

(Consider `.Warning()` since this shouldn't be fatal)

config/snapshot/config_snapshot.go
```diff
@@ -29,130 +28,67 @@ import (
...
-	_, err := cs.update(ctx, client, configPath, &cond)
-	switch {
-	case err == nil:
-		// Configuration changed
-		return true, update(cs)
-	case !gcs.IsPreconditionFailed(err):
-		return false, err
+func updateHash(ctx context.Context, client gcs.ConditionalClient, configPath gcs.Path, cond *storage.Conditions) (ConfigSnapshot, error) {
+	cs := ConfigSnapshot{}
+	if cond != nil {
```
fejta, line 86: `client.If(nil, nil)` will work as expected

config/snapshot/config_snapshot.go
```diff
@@ -29,130 +28,67 @@ import (
...
 
-// Config is a mapped representation of the config at a particular point in time
-type Config struct {
-	dashboards map[string]*configpb.Dashboard
-	groups     map[string]*configpb.TestGroup
-	attrs      storage.ReaderObjectAttrs
-	lock       sync.RWMutex
+// ConfigSnapshot is a mapped representation of the config at a particular point in time.
+// Not concurrency-safe; meant for reading only
+type ConfigSnapshot struct {
```
fejta, line 33: consider `snapshot.Config` rather than `snapshot.ConfigSnapshot`

config/snapshot/config_snapshot.go
```diff
@@ -29,130 +28,67 @@ import (
...
-	cond.GenerationNotMatch = cs.attrs.Generation
-	_, err := cs.update(ctx, client, configPath, &cond)
-	switch {
-	case err == nil:
-		// Configuration changed
-		return true, update(cs)
-	case !gcs.IsPreconditionFailed(err):
-		return false, err
+func updateHash(ctx context.Context, client gcs.ConditionalClient, configPath gcs.Path, cond *storage.Conditions) (ConfigSnapshot, error) {
+	cs := ConfigSnapshot{}
```
fejta, line 85: `var cs ConfigSnapshot` -- you're not initializing anything

config/snapshot/config_snapshot.go
```diff
@@ -29,130 +28,67 @@ import (
...
-	var cond storage.Conditions
-	cond.GenerationNotMatch = cs.attrs.Generation
-	_, err := cs.update(ctx, client, configPath, &cond)
-	switch {
-	case err == nil:
-		// Configuration changed
-		return true, update(cs)
-	case !gcs.IsPreconditionFailed(err):
-		return false, err
+func updateHash(ctx context.Context, client gcs.ConditionalClient, configPath gcs.Path, cond *storage.Conditions) (ConfigSnapshot, error) {
```
fejta, line 84: `(*ConfigSnapshot, error)` seems more correct here, this pattern is more common when things return an error.

config/snapshot/config_snapshot.go
```diff
@@ -29,130 +28,67 @@ import (
...
 }
 
-// Dashboard returns a dashboard's configuration
-func (cs *Config) Dashboard(name string) *configpb.Dashboard {
-	if cs == nil {
-		return nil
+// Observe reads the config at configPath and return a ConfigSnapshot initially and, if a ticker is supplied, when the config changes
+// Returns an error instead if there is no file at configPath
+func Observe(ctx context.Context, log logrus.FieldLogger, client gcs.ConditionalClient, configPath gcs.Path, ticker <-chan time.Time) (<-chan ConfigSnapshot, error) {
+	ch := make(chan ConfigSnapshot)
```
fejta, line 42: Any reason you are not doing `chan *ConfigSnapshot`?

config/snapshot/config_snapshot.go
```diff
@@ -29,130 +28,67 @@ import (
...
 	}
-	return false, nil
-}
-
-func (cs *Config) update(ctx context.Context, client gcs.ConditionalClient, configPath gcs.Path, cond *storage.Conditions) (*storage.ReaderObjectAttrs, error) {
-	cfg, attrs, err := fetchConfig(ctx, client.If(cond, cond), configPath)
+	cfg, attrs, err := fetchConfig(ctx, client, configPath)
 	if err != nil {
-		return nil, err
+		return cs, err
```
fejta, line 91: `nil, err`

config/snapshot/config_snapshot.go
chases2: Didn't need it to be nil-able, and I don't know what other meaningful difference there is between "snapshot.Config, a struct of pointers" and "*snapshot.Config, a pointer to a struct of pointers".

I'm likely to switch this if I switch the signature for `updateHash`; being consistent throughout the code is a pretty good reason.

config/snapshot/config_snapshot.go
chases2: Now logs a warning and does not return

config/snapshot/config_snapshot.go
chases2: Changed to "Warning" (no f)

fejta (approved):

PR #810, opened 2022-02-08
Implement custom evaluator rules for junit results

fejta:


pkg/updater/eval.go
```diff
@@ -0,0 +1,231 @@
...
+	if e := jr.Result.Errored; e != nil {
+		options = append(options, *e)
+	}
+	if f := jr.Result.Failure; f != nil {
+		options = append(options, *f)
+	}
+	return options
+}
+
+func evalProperties(rule *evalpb.Rule, testResult TestResult) *tspb.TestStatus {
```
michelle192837, line 141: nit: Might be nice to split this into multiple functions.

michelle192837 (approved):
/hold

Can you add a quick blurb about this to the configuration docs too? (prob. https://github.com/GoogleCloudPlatform/testgrid/blob/master/config.md#advanced-configuration?)

pkg/updater/eval.go
fejta: What kind of split are you imagining?

fejta (commented):
> Can you add a quick blurb about this to the configuration docs too?

Isn't that just going to encourage people to use this functionality? Are you sure we want that?

"Your database shows the testcase passed but because of some random string testgrid decides to show it as failing" seems like an extremely questionable design...

pkg/updater/eval.go
michelle192837: Was thinking to pull out a function per type (property, test result field, test result error field), but this isn't that long a function so not actually sure it's better for readability/testability.

michelle192837:
> > Can you add a quick blurb about this to the configuration docs too?
> 
> Isn't that just going to encourage people to use this functionality? Are you sure we want that?
> 
> "Your database shows the testcase passed but because of some random string testgrid decides to show it as failing" seems like an extremely questionable design...

I think it's a bit clearer if it's custom evaluator rules since that will be present in the config, but good point. Not documenting for now seems fine.

PR #825, opened 2022-02-11
Config snapshot adjustments

chases2:
- Config Snapshot test now detects if it "does not snapshot when generation match" correctly 
- Deletes unused/duplicate code. This wound up being shockingly little cut, since the "configSnapshot" has already been removed.

/assign @fejta

### DIFF
--- a/pkg/updater/updater.go
+++ b/pkg/updater/updater.go
@@ -34,6 +34,7 @@ import (
 
 	"cloud.google.com/go/storage"
 	"github.com/GoogleCloudPlatform/testgrid/config"
+	"github.com/GoogleCloudPlatform/testgrid/config/snapshot"
 	"github.com/GoogleCloudPlatform/testgrid/internal/result"
 	configpb "github.com/GoogleCloudPlatform/testgrid/pb/config"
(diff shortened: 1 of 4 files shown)


pkg/updater/updater.go
```diff
@@ -285,46 +312,24 @@ func Update(parent context.Context, client gcs.ConditionalClient, mets *Metrics,
 			log.Debug("Started fixers on current test groups")
 		}
 		fixAll()
-		cond := storage.Conditions{GenerationNotMatch: cfgAttrs.Generation}
-		opener := client.If(&cond, &cond)
-		ticker := time.NewTicker(time.Minute) // TODO(fejta): subscribe to notifications
```
fejta, line 290: Why pull this out? It seems fine to me to continue doing both the occasional logging and the occasional cfg snapshot update inside the same go routine?

chases2: It seemed clearer to have each goroutine doing one thing; one for logging and one for config updating. I'm fine with moving it back.

fejta: Fair enough, either way is good with me.

fejta (approved):
/hold

PR #832, opened 2022-02-24
Fix JUnit struct definitions to follow https://llg.cubic.org/docs/junit/

chizhg:
The change does not modify any of TestGrid logic, but to make sure the JUnit struct definitions follow the https://llg.cubic.org/docs/junit/ standard, so that the XML files can be marshaled & unmarshaled correctly when other projects import TestGrid as a library and use these structs.

/cc @chases2 @michelle192837


chizhg:
@michelle192837 @chases2 could you take a look at this PR when you have time? Thanks!

michelle192837 (approved):
Looks good to me, thanks Chi!

PR #843, opened 2022-03-18
Add tab state support to summarizer pubsub fixer

fejta:


pkg/summarizer/pubsub.go
```diff
@@ -98,14 +102,17 @@ func processGCSNotifications(ctx context.Context, log logrus.FieldLogger, q *con
 }
 
 // Return the test group associated with the notification.
```
chases2, line 104: nit: "test group/dashboard"

chases2 (approved):
/hold

pkg/summarizer/pubsub.go
fejta: Done

chases2 (approved):

PR #845, opened 2022-03-18
Ignore poorly named objects

fejta:


pkg/pubsub/pubsub.go
```diff
@@ -119,6 +120,15 @@ func (ra realAcker) Nack(m *pubsub.Message) {
 func sendToReceivers(ctx context.Context, log logrus.FieldLogger, send Sender, receivers chan<- *Notification, result acker) error {
 	return send(ctx, func(ctx context.Context, msg *pubsub.Message) {
 		bucket, obj := msg.Attributes[keyBucket], msg.Attributes[keyObject]
+		if strings.Contains(obj, "#") {
```
chases2, line 123: Can the Notification stream send notifications for versioned objects? If it did, I would imagine the version indicator would be given in the message somewhere, and not in the objectId, right?

chases2 (approved):

pkg/pubsub/pubsub.go
fejta: The generation id has its own field: https://cloud.google.com/storage/docs/pubsub-notifications#attributes

PR #852, opened 2022-03-21
Write knative tab state

fejta:


chases2 (approved):

fejta:
/hold cancel

We didn't actually deploy anything last time...

PR #853, opened 2022-03-21
Script to sync workload identity bindings

fejta:


cluster/bind-service-accounts.sh
```diff
@@ -0,0 +1,114 @@
...
+      echo "NOOP: $membermsg has $rolemsg access to $gsa"
+      continue
+    fi
+
+    case "$add" in
+      y*|Y*)
+      (
+        set -o xtrace
+        "$gcloud" iam service-accounts \
+          --project "$gsa_project" \
```
chases2, line 57: I thought `${gsa_project}` was preferred in bash?

chases2 (approved):
/hold

cluster/bind-service-accounts.sh
fejta: Technically correct: https://google.github.io/styleguide/shellguide.html#quoting but I think shellcheck has more sane guidelines: https://github.com/koalaman/shellcheck I will happily follow it here if someone sets it up

fejta: https://github.com/kubernetes/kubernetes/blob/master/hack/verify-shellcheck.sh

Issue #858, opened 2022-03-22
Enable pubsub for knative summarizer


fejta:
https://github.com/GoogleCloudPlatform/testgrid/blob/331236c91d64e47a0d64dfb4fed00d2062c7aad4/cluster/setup-testgrid-pubsub.sh#L33-L56

https://github.com/GoogleCloudPlatform/testgrid/blob/331236c91d64e47a0d64dfb4fed00d2062c7aad4/cluster/prod/knative/summarizer.yaml#L26-L31

https://github.com/GoogleCloudPlatform/testgrid/blob/331236c91d64e47a0d64dfb4fed00d2062c7aad4/cluster/prod/summarizer.yaml#L43-L44

oh cool note that there is https://github.com/GoogleCloudPlatform/testgrid/blob/master/cluster/prod/knative/setup-pubsub.sh

/assign @chases2

PR #860, opened 2022-03-22
Add pubsub and persistent queue for knative summarizer

chases2:
Splits the "anything" topic into two: one for test group state and one for tab state

/assign @fejta @chaodaiG 
xref #858

### DIFF
--- a/cluster/prod/knative/setup-testgrid-pubsub.sh
+++ b/cluster/prod/knative/setup-testgrid-pubsub.sh
@@ -1,0 +1,49 @@
+#!/usr/bin/env bash
+# Copyright 2022 The TestGrid Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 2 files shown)


cluster/prod/knative/setup-testgrid-pubsub.sh
```diff
@@ -0,0 +1,49 @@
...
+
+    log "$create_sub" -t "${topic_prefix}-grid" -b "$bot" -p "${project}" "$group_sub"
+    log "$create_sub" -t "${topic_prefix}-tabs" -b "$bot" -p "${project}" "$tab_sub"
+}
+
+project=knative-tests
+bucket=gs://knative-own-testgrid
+topic_prefix="projects/$project/topics/knative-own-testgrid"
+group_sub=test-group-updates
+tab_sub=tab-updates
```
fejta, line 47: FYI this will get deleted since nothing is consuming it.

consider spinning up a second summarizer that reads from this subscription just so we keep it around.

fejta (approved):

PR #861, opened 2022-03-22
Consume test-group pubsub subscription with tabulator

fejta:


cluster/canary/summarizer.yaml
```diff
@@ -43,8 +43,9 @@ spec:
         - --confirm
         - --grid-path=grid
         - --json-logs
-        - --persist-queue=gs://k8s-testgrid-canary/queue/summarizer.json
```
chases2, line 46: nit: Why remove the persist queue? Maybe we'll want it to debug why a summary became super slow or something.

chases2 (approved):
/hold unhold at will

cluster/canary/summarizer.yaml
fejta: See the second deployment below it. This deployment doesn't do anything anymore. It only exists to keep the pubsub subscription alive :) 

I'll delete it soon.

And there's no need to persist the queue if it isn't getting any pubsub updates. It already orders the queue based on the last update time on startup.

PR #868, opened 2022-03-24
Add option to override build ID with config value.

michelle192837:
Reference b/222333380 for details.

I don't plan to add more override options until Tabulator is up and running (since I think we'll have better options for how primary grouping/overrides work), but adding this to unblock some Prow users since I had an in-progress PR already.


pb/config/config.proto
```diff
@@ -328,7 +330,7 @@ message TestGroup {
   // These can be substituted into LinkTemplates.
   string user_property = 56;
 
-  reserved 58,59;
+  reserved 57,58,59;
```
fejta, line 333: Does 57-59 work? Do we care? (maybe for next PR)

michelle192837: Good question, I believe comma-separated or range can be mix-matched (https://developers.google.com/protocol-buffers/docs/proto3#reserved). Happy to do a range instead if that reads clearer.

PR #877, opened 2022-03-28
Bump the max size of junit xml(s)

dims:
With the json output switch, we end up with larger output for each line,
which in turn has bumped up the size of the junit xml(s). Please see for
example:

https://testgrid.k8s.io/sig-release-master-blocking#integration-master&width=20

We end up with a message in test grid that xml file is too large.

Signed-off-by: Davanum Srinivas <davanum@gmail.com>

### DIFF
--- a/util/gcs/read.go
+++ b/util/gcs/read.go
@@ -413,7 +413,7 @@ type SuitesMeta struct {
 }
 
 const (
-	maxSize int64 = 100e6 // 100 million, coarce to int not float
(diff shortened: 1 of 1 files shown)


fejta:
Please shrink the size of your XML files

fejta:
Shoot for an xml file of 1MB or so

dims:
@fejta ack. doing so now in https://github.com/kubernetes/kubernetes/pull/109112

PR #887, opened 2022-04-05
Add processing to tabulator

chases2:
Allows processing to be done between tab states. Actually useful processing to be implemented.

Adds a `copy-only` flag to suppress this behavior.

### DIFF
--- a/cmd/tabulator/main.go
+++ b/cmd/tabulator/main.go
@@ -39,6 +39,7 @@ type options struct {
 	persistQueue       gcs.Path
 	creds              string
 	confirm            bool
+	filter             bool
 	concurrency        int
 	wait               time.Duration
 	gridPathPrefix     string
@@ -70,6 +71,7 @@ func gatherOptions() options {
 	flag.Var(&o.persistQueue, "persist-queue", "Load previous queue state from gs://path/to/queue-state.json and regularly save to it thereafter")
 	flag.StringVar(&o.creds, "gcp-service-account", "", "/path/to/gcp/creds (use local creds if empty)")
 	flag.BoolVar(&o.confirm, "confirm", false, "Upload data if set")
+	flag.BoolVar(&o.filter, "filter", false, "Filters out data according to the tab's configuration")
 	flag.IntVar(&o.concurrency, "concurrency", 0, "Manually define the number of groups to concurrently update if non-zero")
 	flag.DurationVar(&o.wait, "wait", 0, "Ensure at least this much time has passed since the last loop (exit if zero).")
 
@@ -138,7 +140,7 @@ func main() {
 
 	mets := tabulator.CreateMetrics(prometheus.NewFactory())
 
-	if err := tabulator.Update(ctx, client, mets, opt.config, opt.concurrency, opt.gridPathPrefix, opt.tabStatePathPrefix, opt.confirm, opt.wait, fixers...); err != nil {
+	if err := tabulator.Update(ctx, client, mets, opt.config, opt.concurrency, opt.gridPathPrefix, opt.tabStatePathPrefix, opt.confirm, opt.filter, opt.wait, fixers...); err != nil {
 		logrus.WithError(err).Error("Could not tabulate")
 	}
 }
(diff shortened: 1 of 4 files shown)


pkg/tabulator/tabstate.go
```diff
@@ -169,8 +174,8 @@ func Update(ctx context.Context, client gcs.ConditionalClient, mets *Metrics, co
 			log.WithFields(logrus.Fields{
 				"from": fromPath.String(),
 				"to":   toPath.String(),
-			}).Info("Copying state")
-			if confirm {
+			}).Info("Calculating state")
+			if copyOnly && confirm {
```
michelle192837, line 178: nit: If confirm = false, think neither copy nor upload should happen?

pkg/tabulator/tabstate.go
```diff
@@ -245,3 +260,46 @@ func TabStatePath(configPath gcs.Path, tabPrefix, dashboardName, tabName string)
 	}
 	return np, nil
 }
+
+func calculateState(ctx context.Context, client gcs.Client, log *logrus.Entry, fromPath, toPath gcs.Path, confirm bool) error {
```
michelle192837, line 264: nit: calculateState doesn't indicate to me that this does uploads, is there a better name to clarify that?

michelle192837 (approved):
/hold

Looks good, couple of nits!

pkg/tabulator/tabstate.go
```diff
@@ -245,3 +260,46 @@ func TabStatePath(configPath gcs.Path, tabPrefix, dashboardName, tabName string)
...
+	buf, err := ioutil.ReadAll(z)
+	if err != nil {
+		return fmt.Errorf("ioutil.ReadAll: %w", err)
+	}
+	var g statepb.Grid
+	if err = proto.Unmarshal(buf, &g); err != nil {
+		return fmt.Errorf("proto.Unmarshal: %w", err)
+	}
+
+	// TODO(chases2): From summarizer: filterGrid()
```
fejta, line 284: IMO this PR is fine but not particularly useful. 

This just demonstrates using the proto and zip packages, which seems unlikely to cause a problem.

The more interesting piece here is to inflate the grid and then deflate it.

AKA convert it into `[]InflatedGrid` and back

https://github.com/GoogleCloudPlatform/testgrid/blob/90494ebb6318b4a135aa950c1d70d087d74a81ab/pkg/updater/inflate.go#L32-L37

One idea might be to call `InflateDropAppend` with a noop column reader:

https://github.com/GoogleCloudPlatform/testgrid/blob/90494ebb6318b4a135aa950c1d70d087d74a81ab/pkg/updater/updater.go#L549

fejta (approved):

cmd/tabulator/main.go
```diff
@@ -70,6 +71,7 @@ func gatherOptions() options {
 	flag.Var(&o.persistQueue, "persist-queue", "Load previous queue state from gs://path/to/queue-state.json and regularly save to it thereafter")
 	flag.StringVar(&o.creds, "gcp-service-account", "", "/path/to/gcp/creds (use local creds if empty)")
 	flag.BoolVar(&o.confirm, "confirm", false, "Upload data if set")
+	flag.BoolVar(&o.copyOnly, "copy-only", true, "Copies grid state to tab state with no transformation")
```
fejta, line 74: Please default this to off until we roll it out to prod.

pkg/tabulator/tabstate.go
chases2: I don't think we'll need to inflate/deflate the grid in order to filter it. The summarizer doesn't.

We will need to inflate/deflate to drop columns later, to determine alerting and to drop empty columns after filtering, but I intend to do that in a later PR.

cmd/tabulator/main.go
chases2: It was defaulted "off", but it wasn't as clear I don't think. Switched the flag around

pkg/tabulator/tabstate.go
chases2: Yes, that's right. There was a bug here my tests didn't catch; thanks!

pkg/tabulator/tabstate.go
chases2: Chose a more vague name, since this function currently "does everything" for a particular tab.

PR #890, opened 2022-04-06
Add filtering to tabulator

chases2:
Allows the tabulator to use base options to remove entire rows that will be filtered out by the UI and summarizer anyway.

/hold (for further testing)

### DIFF
--- a/pkg/tabulator/filter.go
+++ b/pkg/tabulator/filter.go
@@ -1,0 +1,91 @@
+/*
+Copyright 2022 The TestGrid Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package tabulator processes test group state into tab state.
+package tabulator
+
+import (
+	"fmt"
+	"net/url"
+	"regexp"
+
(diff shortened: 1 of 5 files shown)


pkg/tabulator/filter.go
```diff
@@ -0,0 +1,91 @@
...
+	statepb "github.com/GoogleCloudPlatform/testgrid/pb/state"
+)
+
+const (
+	includeFilter = "include-filter-by-regex"
+	excludeFilter = "exclude-filter-by-regex"
+)
+
+// filterGrid filters the grid to rows matching the include/exclude list in the base options
+func filterGrid(baseOptions string, rows []*statepb.Row) ([]*statepb.Row, error) {
```
fejta, line 34: Is this the same code as the summarizer? If so I'd just move this function (and test code) into this package and then call it from the summarizer, which already imports this package.

https://github.com/GoogleCloudPlatform/testgrid/blob/c39f9b3665f70bad137d27be89e495dba1022755/pkg/summarizer/summary.go#L44

My main concern is that I think we should continue calling filterGrid from the summarizer for a bit longer after we release this, aka we should still do:

https://github.com/GoogleCloudPlatform/testgrid/blob/c39f9b3665f70bad137d27be89e495dba1022755/pkg/summarizer/summary.go#L604

Happy to have a single `FilterGrid` that's called by both summarizer and tabulator though (also fine to keep the copy)

(same thing for includeRows, excludeRows)

fejta (approved):

pkg/tabulator/filter.go
chases2: I plan to continue filtering in the summarizer for now. Removing that code is an out-of-band change. (See https://github.com/GoogleCloudPlatform/testgrid/pull/891 for removing that functionality by flag)

Given that this is the new home for this code, a little copying seemed better than a little importing. When dropping empty columns (the next TODO), it's likely I'll want to refactor this code without mucking up the summarizer's current functionality.

PR #891, opened 2022-04-06
Add skip-filter flag for summarizer

chases2:
Adds a flag that allows you to turn off regex filtering (if it's already being done upstream by the tabulator.

### DIFF
--- a/pkg/summarizer/summary.go
+++ b/pkg/summarizer/summary.go
@@ -89,7 +89,7 @@ type Fixer func(context.Context, *config.DashboardQueue) error
 // Will use concurrency go routines to update dashboards in parallel.
 // Setting dashboard will limit update to this dashboard.
 // Will write summary proto when confirm is set.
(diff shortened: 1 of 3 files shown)


pkg/summarizer/summary.go
```diff
@@ -750,39 +749,37 @@ func firstFilled(values ...int32) int {
...
+func filterGrid(baseOptions string, rows []*statepb.Row, recent int, useRegexFilter bool) ([]*statepb.Row, error) {
 
 	rows = recentRows(rows, recent)
 
 	rows = filterMethods(rows)
 
-	for _, include := range vals[includeFilter] {
-		if rows, err = includeRows(rows, include); err != nil {
-			return nil, fmt.Errorf("bad %s=%s: %v", includeFilter, include, err)
+	if useRegexFilter {
```
michelle192837, line 764: nit: Maybe `if !useRegexFilter { return rows, nil }` to limit what's inside the conditional?

michelle192837 (approved):
/hold

Unhold at will

pkg/summarizer/summary.go
chases2: Good idea; done

PR #899, opened 2022-04-08
Enable filter in canary

chases2:


michelle192837 (approved):

fejta:
(For next time, I'd probably try and do it on a handful of groups before doing it on every group [to minimize blast radius if something fails])

PR #905, opened 2022-04-08
Drop empty columns after filtering by regex

chases2:
TODO: 
- [x] Gate this feature behind a flag also

### DIFF
--- a/pkg/tabulator/tabstate.go
+++ b/pkg/tabulator/tabstate.go
@@ -18,12 +18,9 @@ limitations under the License.
 package tabulator
 
 import (
-	"bytes"
-	"compress/zlib"
 	"context"
 	"errors"
 	"fmt"
-	"io/ioutil"
 	"net/url"
 	"path"
 	"sync"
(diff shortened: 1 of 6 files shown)


pkg/tabulator/tabstate.go
```diff
@@ -276,50 +269,70 @@ func TabStatePath(configPath gcs.Path, tabPrefix, dashboardName, tabName string)
...
+	for index, col := range grid {
+		drop := true
+		for _, cell := range col.Cells {
+			if cell.Result != test_status.TestStatus_NO_RESULT {
+				drop = false
+				break
+			}
+		}
+		if drop {
+			result = append(result, grid[lastIndex:index]...)
```
michelle192837, line 328: I might be missing something, but is there a reason this appends multiple things at once instead of appending a single column at a time? (I think the latter might be more readable, e.g. for each col, append to result if it's non-empty).

michelle192837 (commented):
Couple of comments, otherwise looks good!

pkg/tabulator/tabstate.go
```diff
@@ -276,50 +269,70 @@ func TabStatePath(configPath gcs.Path, tabPrefix, dashboardName, tabName string)
 	return np, nil
 }
 
-func tabulate(ctx context.Context, client gcs.Client, cfg *configpb.DashboardTab, testGroupPath, tabStatePath gcs.Path, confirm bool) error {
-	r, _, err := client.Open(ctx, testGroupPath)
+func tabulate(ctx context.Context, client gcs.Client, cfg *configpb.DashboardTab, testGroupPath, tabStatePath gcs.Path, confirm, dropEmptyCols bool) error {
+	rawGrid, _, err := gcs.DownloadGrid(ctx, client, testGroupPath)
 	if err != nil {
-		return fmt.Errorf("client.Open(%s): %w", testGroupPath.String(), err)
+		return fmt.Errorf("downloadGrid: %w", err)
```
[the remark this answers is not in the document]
chases2: Done

pkg/tabulator/tabstate.go
chases2: Fewer append() calls, but more complexity and local variables. Not sure it's worth it. WDYT of this version?

michelle192837: Oh that's even simpler than I was thinking, looks great! ^^

michelle192837 (approved):

PR #928, opened 2022-04-21
Calculate column stats when broken column threshold is specified.

michelle192837:
This helps flag broken columns (in tab state, where we have broken column thresholds specified), and will later help calculate column alerts, and exclude broken columns from row alerts.


pkg/updater/updater.go
```diff
@@ -1082,6 +1082,9 @@ func ConstructGrid(log logrus.FieldLogger, cols []InflatedColumn, issues map[str
 	}
 
 	for _, col := range cols {
+		if col.Column != nil {
```
fejta, line 1085: Is it worth putting this new behavior behind a flag?

fejta (approved):
/hold

pkg/updater/updater.go
michelle192837: I was debating on it, but yeah might as well!

fejta: Is this a new field? If so then maybe not so important. Does it even appear in the UI?

fejta: But still IMO good practice!~

michelle192837: New field, doesn't appear in the UI yet (though maybe it'll be a nice FixIt item to flag columns as broken, not quite trivial due to how column headers get combined rn). Given this is a shared updater function, would it make sense to flag this in tabulator:main + use `brokenThreshold > 0.0` explicitly here?

fejta: Sure, the goal here is mostly to design the failure, maybe when we encounter "this is for mysterious reasons causing testgrid to crash" or "all the updaters are surprisingly slow after this push" that we have a cheap way to turn this functionality off while we investigate.

pkg/updater/updater.go
```diff
@@ -1419,6 +1422,42 @@ func alertInfo(failures int32, msg, cellID, latestCellID string, userProperties
...
+		default:
+			// blank cell or unrecognized status, do nothing
+		}
+	}
+	return &statepb.Stats{
+		FailCount:  fails,
+		PassCount:  passes,
+		TotalCount: total,
+		Pending:    pending,
+		Broken:     float32(fails)/float32(total) > brokenThreshold,
```
fejta, line 1457: if total is 0 this will blow up, for example

michelle192837: Daaaaaangit.

fejta (approved):

PR #931, opened 2022-04-21
Validate names only contain allowed characters.

michelle192837:


config/config.go
```diff
@@ -208,9 +208,15 @@ func validateReferencesExist(c *configpb.Configuration) error {
 	return mErr
 }
 
-// validateName validates an entity name is non-empty and contains no prefix that overlaps with a
-// TestGrid file prefix, post-normalization.
+// TODO(michelle192837): Remove '/' and '–' from this regex.
+var nameRegex = regexp.MustCompile("^[a-zA-Z0-9_.~<>()|\\[\\]\",@/ –-]+$")
```
fejta, line 212: Backtick is another option

fejta (approved):

config/config.go
michelle192837: Thanks! Added at https://github.com/GoogleCloudPlatform/testgrid/pull/932

PR #939, opened 2022-05-05
Set Tabulator cache settings to "verified cache only"

chases2:
- Fixes a tabulator bug where old cache values can be served to critical service components
- Fixes (presumably) a similar bug with queue persistence files
- Hopefully makes the "no-cache" string a little less magic. It's not intuitive that no-cache does cache, but returns verified cache entries.

### DIFF
--- a/util/gcs/gcs.go
+++ b/util/gcs/gcs.go
@@ -168,10 +168,13 @@ func calcCRC(buf []byte) uint32 {
 }
 
 const (
-	// DefaultACL for this upload
+	// Use DefaultACL for this upload.
 	DefaultACL = false
-	// PublicRead ACL for this upload.
+	// Use PublicRead ACL for this upload.
 	PublicRead = true
+	// May cache, but only after verifying.
+	// See https://cloud.google.com/storage/docs/metadata#cache-control
+	NoCache = "no-cache"
 )
 
 // Upload writes bytes to the specified Path by converting the client and path into an ObjectHandle.
--- a/pkg/merger/merger.go
+++ b/pkg/merger/merger.go
@@ -198,7 +198,7 @@ func MergeAndUpdate(ctx context.Context, client mergeClient, mets *Metrics, list
 		return result, fmt.Errorf("can't marshal merged proto: %w", err)
 	}
 
-	if _, err := client.Upload(ctx, *list.Path, buf, gcs.DefaultACL, "no-cache"); err != nil {
+	if _, err := client.Upload(ctx, *list.Path, buf, gcs.DefaultACL, gcs.NoCache); err != nil {
 		finish.Fail()
 		return result, fmt.Errorf("can't upload merged proto to %s: %w", list.Path, err)
 	}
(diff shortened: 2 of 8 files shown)


util/gcs/gcs.go
```diff
@@ -172,6 +172,9 @@ const (
 	DefaultACL = false
 	// PublicRead ACL for this upload.
 	PublicRead = true
+	// Default cache settings for this upload.
+	// Will cache, but only after verifying. See https://cloud.google.com/storage/docs/metadata#cache-control
+	DefaultCache = "no-cache"
```
fejta, line 177: IMO `NoCache` is a better name for this. GCS does not support a customizable default cache setting.

`DefaultACL` is a specific GCS feature -- aka the set of ACLs to apply when not explicitly set by the upload. It is configurable too, see:

* `gsutil defacl`
* https://cloud.google.com/storage/docs/access-control/lists#default

fejta (approved):
/hold

util/gcs/gcs.go
chases2: Ah, I see. Mixing these two senses of "default" is no good.

I don't like "NoCache" either, though; it's misleading. Might go with "CheckCache", "ValidateCache", or something to reflect how GCS acts.

util/gcs/gcs.go
```diff
@@ -172,6 +172,9 @@ const (
 	DefaultACL = false
 	// PublicRead ACL for this upload.
 	PublicRead = true
+	// May cache, but only after verifying.
+	// See https://cloud.google.com/storage/docs/metadata#cache-control
+	ValidateCache = "no-cache"
```
fejta, line 177: Not a huge deal, but I continue to think `NoCache` is a better name for using the `no-cache` value in this header.

What is the motivation for creating an abstraction layer here? It just seems to distance people from understanding what is actually happening.

AKA I think it is better for people to think concretely:
* "I want to upload this file to GCS with a `Cache-Control: no-cache` header" -- and understand how GCS will work with this

Than to have some abstract idea:
* "I want this file to be validated before using the cached version" or something and hope that this abstract understanding matches what GCS will do.

The reality is that the only way for them to be sure about this is to understand a) that `ValidateCache` means `no-cache` and b) how GCS behaves with `no-cache`, at which point this abstraction becomes useless.

Also, using validate seems to conflate with the way this header can use `revalidate`

chases2: Abstraction isn't what I'm interested in here, and I don't want to abstract in a way that makes it less obvious what this does. (Not being able to mistype "nocache" is nice, but not my goal)
 
What I'm want is for the argument to either describe what it does ("Validate caches before serving them") or, barring that, be strange enough that I'll hit F12, see the documentation link, and remember that "no-cache" is complicated when it doesn't appear complicated.
 
I still think that "no-cache" is not an ideal name for that _header_, although I'd have to take that up with the [RFC](https://datatracker.ietf.org/doc/html/rfc7234#section-5.2).

PR #942, opened 2022-05-11
Refactor tabulator algorithm to tabulate by test group

chases2:
Generally seems to run faster at various concurrency levels, while using similar amounts of memory.

/assign @fejta

### DIFF
--- a/pkg/tabulator/tabstate.go
+++ b/pkg/tabulator/tabstate.go
@@ -56,14 +56,32 @@ func CreateMetrics(factory metrics.Factory) *Metrics {
 	}
 }
 
-// Fixer should adjust the dashboard queue until the context expires.
-type Fixer func(context.Context, *config.DashboardQueue) error
+type view struct {
+	dashboard *configpb.Dashboard
+	tab       *configpb.DashboardTab
+}
+
+func getViews(cfg *snapshot.Config) map[string][]view {
+	groupToTabs := make(map[string][]view, len(cfg.Groups))
+
+	for _, dashboard := range cfg.Dashboards {
+		for _, tab := range dashboard.DashboardTab {
+			g := tab.TestGroupName
+			groupToTabs[g] = append(groupToTabs[g], view{dashboard, tab})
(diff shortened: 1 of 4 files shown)


pkg/tabulator/tabstate.go
```diff
@@ -198,8 +217,7 @@ func Update(ctx context.Context, client gcs.ConditionalClient, mets *Metrics, co
 				}
 			}
 			if filter {
-				groupCfg := cfg.Groups[tab.GetTestGroupName()]
-				err := createTabState(ctx, log, client, tab, groupCfg, *fromPath, *toPath, confirm, dropEmptyCols, calculateStats, useTabAlertSettings)
+				err := createTabState(ctx, log, client, view.tab, group, *fromPath, *toPath, confirm, dropEmptyCols, calculateStats, useTabAlertSettings)
```
fejta, line 220: Seems like we should refactor group out of this function.

Instead of downloading and inflating the group state n-view times we can do this once, and then send a copy of the inflated state to all the views for further processing.

(but we'll have to be careful to do a deep copy or whatnot as maps and slices are passed as references)

pkg/tabulator/tabstate.go
```diff
@@ -214,44 +232,44 @@ func Update(ctx context.Context, client gcs.ConditionalClient, mets *Metrics, co
 	for i := 0; i < concurrency; i++ {
 		go func() {
 			defer wg.Done()
-			for dashName := range dashboardNames {
+			for group := range channel {
 				lock.Lock()
-				start := active.Add(dashName)
+				start := active.Add(group.Name)
 				if !start {
```
fejta, line 238: We can now have two go routines concurrently updating different parts of the same dashboard... 

What protects the one routine from updating the dashboard update done by another? The fact that the tab state is stored on a tab basis, unlike the summarizer which is stored on a dashboard basis (something to keep in mind in case a future refactor violates this hidden design dependency).

fejta (approved):

PR #947, opened 2022-05-12
Set filter by regex as default behavior for tabulator

chases2:
I don't think this is worth maintaining as a mode of operation

### DIFF
--- a/cmd/tabulator/main.go
+++ b/cmd/tabulator/main.go
@@ -77,8 +77,8 @@ func gatherOptions() options {
 	flag.StringVar(&o.creds, "gcp-service-account", "", "/path/to/gcp/creds (use local creds if empty)")
 	flag.BoolVar(&o.confirm, "confirm", false, "Upload data if set")
 	flag.Var(&o.groups, "group", "Only update named test group if set (repeateable)")
-	flag.BoolVar(&o.filter, "filter", false, "Filters out data according to the tab's configuration") // TODO(chases2): Switch to default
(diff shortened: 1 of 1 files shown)


fejta (approved):

fejta (approved):
/hold

cmd/tabulator/main.go
```diff
@@ -77,8 +81,8 @@ func gatherOptions() options {
 	flag.StringVar(&o.creds, "gcp-service-account", "", "/path/to/gcp/creds (use local creds if empty)")
 	flag.BoolVar(&o.confirm, "confirm", false, "Upload data if set")
 	flag.Var(&o.groups, "group", "Only update named test group if set (repeateable)")
-	flag.BoolVar(&o.filter, "filter", false, "Filters out data according to the tab's configuration") // TODO(chases2): Switch to default
-	flag.BoolVar(&o.dropEmptyCols, "filter-columns", false, "Drops empty columns after filtering")    // TODO(chases2): Enable, then remove flag
+	flag.BoolVar(&o.filter, "filter", false, "Deprecated; will be deleted after June 30 2022")     // TODO: Delete
```
fejta, line 84: We should stop using it before we deprecate it.

https://github.com/GoogleCloudPlatform/testgrid/blob/3cb08bb7a224016bc96762b6e45dd6fe17362d0a/cluster/canary/tabulator.yaml#L38

I would suggest:

1) default the flag to desired behavior,
2) remove usage of the flag from production
3) remove the flag from the code

chases2: Sounds good to me
/hold cancel

PR #948, opened 2022-05-13
Refactor Tabulator to read and write in separate pools

chases2:
Follow-up to https://github.com/GoogleCloudPlatform/testgrid/pull/942

### DIFF
--- a/pkg/tabulator/tabstate.go
+++ b/pkg/tabulator/tabstate.go
@@ -26,8 +26,8 @@ import (
 	"sync"
 	"time"
 
-	"bitbucket.org/creachadair/stringset"
 	"github.com/sirupsen/logrus"
+	"google.golang.org/protobuf/proto"
 
 	"github.com/GoogleCloudPlatform/testgrid/config"
 	"github.com/GoogleCloudPlatform/testgrid/config/snapshot"
@@ -40,6 +40,7 @@ import (
 )
 
 const componentName = "tabulator"
+const writeTimeout = 10 * time.Minute
 
 // Metrics holds metrics relevant to this controller.
 type Metrics struct {
@@ -55,18 +56,24 @@ func CreateMetrics(factory metrics.Factory) *Metrics {
 	}
 }
 
-type view struct {
+type writeTask struct {
 	dashboard *configpb.Dashboard
 	tab       *configpb.DashboardTab
+	group     *configpb.TestGroup
+	data      *statepb.Grid //TODO(chases2): change to inflatedColumns (and additional data) once "filter-columns" is used everywhere
 }
 
-func getViews(cfg *snapshot.Config) map[string][]view {
-	groupToTabs := make(map[string][]view, len(cfg.Groups))
+func mapTasks(cfg *snapshot.Config) map[string][]writeTask {
+	groupToTabs := make(map[string][]writeTask, len(cfg.Groups))
 
 	for _, dashboard := range cfg.Dashboards {
 		for _, tab := range dashboard.DashboardTab {
 			g := tab.TestGroupName
-			groupToTabs[g] = append(groupToTabs[g], view{dashboard, tab})
+			groupToTabs[g] = append(groupToTabs[g], writeTask{
+				dashboard: dashboard,
+				tab:       tab,
+				group:     cfg.Groups[dashboard.Name],
+			})
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


chases2:
/assign @fejta 
Please take a look when you are able!

pkg/tabulator/tabstate.go
```diff
@@ -183,80 +197,133 @@ func Update(ctx context.Context, client gcs.ConditionalClient, mets *Metrics, co
...
+			q.Fix(group.Name, time.Now().Add(freq/10), false)
+			return nil
+		}
+
+		fromPath, err := updater.TestGroupPath(configPath, gridPathPrefix, group.Name)
+		if err != nil {
+			return fmt.Errorf("can't make tg path %q: %w", group.Name, err)
+		}
+
+		log.WithFields(logrus.Fields{
```
fejta, line 221: `log.WithField("from", fromPath)` as you're only using a single field.

pkg/tabulator/tabstate.go
```diff
@@ -183,80 +197,133 @@ func Update(ctx context.Context, client gcs.ConditionalClient, mets *Metrics, co
...
-			}).Info("Calculating state")
-
-			err = createTabState(ctx, log, client, view.tab, group, *fromPath, *toPath, confirm, dropEmptyCols, calculateStats, useTabAlertSettings)
-			if err != nil {
-				return fmt.Errorf("can't calculate state: %w", err)
-			}
+			out := view
+			out.data = proto.Clone(grid).(*statepb.Grid)
+			out.done = wg.Done
+			tabs <- out
```
fejta, line 238: Shouldn't this pay attention to when the context expires?

pkg/tabulator/tabstate.go
```diff
@@ -183,80 +197,133 @@ func Update(ctx context.Context, client gcs.ConditionalClient, mets *Metrics, co
...
-				"to":   toPath.String(),
-			}).Info("Calculating state")
-
-			err = createTabState(ctx, log, client, view.tab, group, *fromPath, *toPath, confirm, dropEmptyCols, calculateStats, useTabAlertSettings)
-			if err != nil {
-				return fmt.Errorf("can't calculate state: %w", err)
-			}
+			out := view
+			out.data = proto.Clone(grid).(*statepb.Grid)
+			out.done = wg.Done
```
fejta, line 237: hmm, are you sure about this? Seems fragile

pkg/tabulator/tabstate.go
```diff
@@ -183,80 +197,133 @@ func Update(ctx context.Context, client gcs.ConditionalClient, mets *Metrics, co
...
-
-				log := log.WithField("group", group.Name)
+			}
+		}()
+	}
+	writeWg.Add(writeConcurrency)
+	for i := 0; i < writeConcurrency; i++ {
+		go func() {
+			defer writeWg.Done()
+			for tab := range tabs {
```
fejta, line 271: Would task be a better name here?

pkg/tabulator/tabstate.go
```diff
@@ -183,80 +197,133 @@ func Update(ctx context.Context, client gcs.ConditionalClient, mets *Metrics, co
...
-				"from": fromPath.String(),
-				"to":   toPath.String(),
-			}).Info("Calculating state")
-
-			err = createTabState(ctx, log, client, view.tab, group, *fromPath, *toPath, confirm, dropEmptyCols, calculateStats, useTabAlertSettings)
-			if err != nil {
-				return fmt.Errorf("can't calculate state: %w", err)
-			}
+			out := view
+			out.data = proto.Clone(grid).(*statepb.Grid)
```
fejta, line 236: Seems a waste of inflate N times? Isn't this what we spend most of our CPU time doing?

fejta (approved):
/hold

pkg/tabulator/tabstate.go
chases2: Not sure what you mean. Deep-copying the result is cheaper than inflating it each time.

pkg/tabulator/tabstate.go
chases2: Good point; renamed

pkg/tabulator/tabstate.go
chases2: Added a skip here

pkg/tabulator/tabstate.go
fejta: This code downloads the proto once and then inflates it N times.

pkg/tabulator/tabstate.go
fejta: WaitGroup seems fine, but passing a wait group to a foreign function via a channel seems fragile and hard to reason about.

fejta: (I want be able to read a single function to validate that `wg.Done()` is called `wg.Add(n)` times)

pkg/tabulator/tabstate.go
chases2: Ah, yes, of course it does. I have no idea why I didn't see that before. I'll look into passing inflated columns or something similar

pkg/tabulator/tabstate.go
fejta: Maybe instead of passing a wg.Done why not have the writer remove the item from the processing set?

chases2: The writer might not be the only one writing a test group; test groups to tabs are one-to-many. The processing set could be dashboard#tabs instead, though. 🤔

fejta: Happy to go with this and wait until there's empirical evidence from debugging a mysterious "why does the tabulator stop doing anything after awhile?" issue after a refactor to conclude that this is a fragile pattern.

pkg/tabulator/tabstate.go
chases2: I can't reasonably fix this until "--filter-columns" is dropped as a feature flag. Will have to address this in a later PR.

pkg/tabulator/tabstate.go
chases2: I thought a bit more about _why_ we want to call back to the reading thread like this at all, and I'm not convinced we need it. These routines are already architected to pass off everything before another reader gets to pass off anything.

Deleting code is a pretty good way to make it simpler, so I would favor that solution

fejta (approved):

PR #953, opened 2022-05-16
Remove filter flag from canary

chases2:
Default is true, so the flag does nothing.

Cannot be migrated to prod until https://github.com/GoogleCloudPlatform/testgrid/pull/951 merges.


fejta (approved):
/hold

PR #961, opened 2022-05-17
Turn on pooled updating in canary.

fejta:
This will concurrently download builds in test groups, instead
of serially download builds in concurrent test groups.


chases2 (approved):
/hold for author to merge

PR #971, opened 2022-05-23
Improve Iter() performance.

fejta:
Channels are much slower than calling a function

```
1m results:
pkg: github.com/GoogleCloudPlatform/testgrid/internal/result
BenchmarkIterSlow-10            78166888                14.90 ns/op
BenchmarkIterFast-10            759703227                1.559 ns/op

10m results:
pkg: github.com/GoogleCloudPlatform/testgrid/internal/result
BenchmarkIterSlow-10                   1        1360214667 ns/op
BenchmarkIterFast-10            728631373                1.590 ns/op
```

### DIFF
--- a/internal/result/results_test.go
+++ b/internal/result/results_test.go
@@ -19,6 +19,7 @@ package result
 import (
 	"context"
 	"fmt"
+	"math/rand"
 	"reflect"
 	"testing"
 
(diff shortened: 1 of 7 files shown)


chases2 (approved):
/hold for other reviewers

internal/result/results_test.go
```diff
@@ -260,3 +253,61 @@ func TestIter(t *testing.T) {
 		})
 	}
 }
+
+const totalResults = 10e6
```
chases2, line 257: nit: To test this between 1 million and 10 million, you modified this constant, right? Consider a table-driven benchmark (https://go.dev/blog/subtests).

Running this at 10m makes it really clear why one is fast and one is slow.

michelle192837 (approved):

internal/result/results_test.go
fejta: https://github.com/GoogleCloudPlatform/testgrid/pull/973

PR #994, opened 2022-05-27
Truncate testgroup state inline

chases2:
Instead of adding a "Truncated" row to indicate truncated columns, simply cut the columns and replace the last cell in each row with an "unknown" marker.


michelle192837 (approved):
/hold

In case you want additional review, this looks good to me though!

fejta (approved):

PR #1012, opened 2022-06-09
add attribute to metrics

ziyulin0:


fejta:
Please `bazel run @io_k8s_repo_infra//hack:update-bazel` (or `:update-deps`)

pkg/merger/merger.go
```diff
@@ -85,15 +87,17 @@ func ParseAndCheck(data []byte) (list MergeList, err error) {
...
 }
 
 // CreateMetrics creates metrics for the Config Merger
 func CreateMetrics(factory metrics.Factory) *Metrics {
 	return &Metrics{
-		Update: factory.NewCyclic(componentName),
-		Fields: factory.NewInt64("config_fields", "Config field usage by name", "component", "field"),
+		Update:       factory.NewCyclic(componentName),
+		Fields:       factory.NewInt64("config_fields", "Config field usage by name", "component", "field"),
+		LastModified: factory.NewInt64("last_modified", "Seconds since shard last modified ", "shard"), // the field name need to be decided
```
michelle192837, line 100: /hold

Until this comment is removed (e.g. the metric name is good enough).

pkg/merger/merger.go
```diff
@@ -206,3 +211,17 @@ func MergeAndUpdate(ctx context.Context, client mergeClient, mets *Metrics, list
...
 	return result, nil
 }
+
+func recordLastModified(attrs *storage.ReaderObjectAttrs, mets *Metrics, source string) {
+	if attrs != nil {
+		lastModified := attrs.LastModified
+		diff := time.Since(lastModified)
+		if mets != nil {
+			mets.LastModified.Set(int64(diff.Seconds()), source)
+			logrus.WithFields(logrus.Fields{
```
chases2, line 221: I would log this information even when metrics are disabled. Info is a good log level.

pkg/merger/merger.go
chases2: "shard" is a good name for a piece of config

pkg/merger/merger_test.go
```diff
@@ -408,6 +408,71 @@ func Test_MergeAndUpdate(t *testing.T) {
...
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			recordLastModified(tc.attrs, tc.mets, tc.source)
+		})
+	}
+}
+
+func Test_RecordLastModified_RecordMetrics(t *testing.T) {
```
chases2, line 445: I would combine this test and the previous test into one table-driven test. The table might be something like

```
name string
attrs *storage.ReaderObjectAttrs
mets *Metrics
expectRecord bool
expectValue int
```

... then you could check all these cases in one test.

pkg/merger/merger_test.go
```diff
@@ -408,6 +408,71 @@ func Test_MergeAndUpdate(t *testing.T) {
...
+	attrs := &storage.ReaderObjectAttrs{
+		LastModified: time.Now(),
+	}
+	mets := &Metrics{
+		LastModified: &fakeMetric,
+	}
+	source := ""
+
+	recordLastModified(attrs, mets, source)
+	if fakeMetric.last != 0 {
```
chases2, line 457: This test relies on the test taking less than one second to execute. Probably a safe bet, but there might be a better way to check that the metric is calculated correctly (and not just a golang zero-value)

```
LastModified: time.Now().Sub(5 * time.Second)
...
if fakeMetric.last < tc.ExpectAtLeast {
     t.Errorf("want at least %d, got %d", tc.ExpectAtLeast, fakeMetric.last)
}
```

pkg/merger/merger_test.go
chases2: https://github.com/GoogleCloudPlatform/testgrid/blob/master/pkg/tabulator/filter_test.go has some examples of these "expect" booleans and expected outputs.

pkg/merger/merger_test.go
```diff
@@ -408,6 +408,71 @@ func Test_MergeAndUpdate(t *testing.T) {
...
+	read bool // Fake implementation not concurrent-safe
+	last int64
+}
+
+func (f *FakeInt64) Name() string {
+	return "FakeInt64"
+}
+
+func (f *FakeInt64) Set(n int64, _ ...string) {
+	if f.read {
```
chases2, line 472: I don't think you ever set f.read to true in these tests, so I'm not sure if this fake works. The test is just checking for a [zero-value](https://go.dev/tour/basics/12) so it's hard to tell.

I might expect a boolean to see if "Set" was called. Something like
```
f.set = true
f.last = n
```
... then your tests can just check if the fake metric was set and what it was set to.

pkg/merger/merger.go
ziyulin0: > "shard" is a good name for a piece of config

Do you mean we should keep the line: `LastModified: factory.NewInt64("last_modified", "Seconds since shard last modified ", "shard"),` without any change?

pkg/merger/merger_test.go
```diff
@@ -408,6 +408,71 @@ func Test_MergeAndUpdate(t *testing.T) {
...
+func Test_RecordLastModified_RecordMetrics(t *testing.T) {
+	var fakeMetric FakeInt64
+
+	attrs := &storage.ReaderObjectAttrs{
+		LastModified: time.Now(),
+	}
+	mets := &Metrics{
+		LastModified: &fakeMetric,
+	}
+	source := ""
```
fejta, line 454: `var source string`

pkg/merger/merger_test.go
```diff
@@ -408,6 +408,71 @@ func Test_MergeAndUpdate(t *testing.T) {
...
+		t.Run(tc.name, func(t *testing.T) {
+			recordLastModified(tc.attrs, tc.mets, tc.source)
+		})
+	}
+}
+
+func Test_RecordLastModified_RecordMetrics(t *testing.T) {
+	var fakeMetric FakeInt64
+
+	attrs := &storage.ReaderObjectAttrs{
```
fejta, line 448: feel free to ignore, but personally I prefer the following style:

```golang

attrs := storage.ReaderObjectAttrs{ 
  // etc
}

recordLastModified(&attrs, etc)
```

AKA declare it as a non-pointer when possible, and use its `&value` , rather than declaring it as a pointer.

pkg/merger/merger_test.go
```diff
@@ -408,6 +408,71 @@ func Test_MergeAndUpdate(t *testing.T) {
...
+	source := ""
+
+	recordLastModified(attrs, mets, source)
+	if fakeMetric.last != 0 {
+		t.Errorf("recordLastModified() fail; got %d, want %d", fakeMetric.last, 0)
+	}
+}
+
+type FakeInt64 struct {
+	read bool // Fake implementation not concurrent-safe
```
fejta, line 463: Seems like you ought to make it concurrent safe (at least if this will be used outside this package).

pkg/merger/merger_test.go
```diff
@@ -408,6 +408,71 @@ func Test_MergeAndUpdate(t *testing.T) {
...
+	}
+	source := ""
+
+	recordLastModified(attrs, mets, source)
+	if fakeMetric.last != 0 {
+		t.Errorf("recordLastModified() fail; got %d, want %d", fakeMetric.last, 0)
+	}
+}
+
+type FakeInt64 struct {
```
fejta, line 462: What other packages are you exporting `FakeInt64` to?

If no one, then please do not export it: `fakeInt64`

pkg/merger/merger_test.go
chases2: I think it would be good to rename this to a "fakeInt64" to make it private.

Keeping things package-scoped until you need them outside the package is generally good, since private -> public is easier than public -> private.

chaodaiG:
I think you need to insert "@com_google_cloud_go_storage//:go_default_library",` after https://github.com/GoogleCloudPlatform/testgrid/blob/56824339d3700064050049c26ccb41ca07e5129d/pkg/merger/BUILD.bazel#L13

pkg/merger/BUILD.bazel
```diff
@@ -11,6 +11,7 @@ go_library(
         "//util/gcs:go_default_library",
         "//util/metrics:go_default_library",
         "@com_github_golang_protobuf//proto:go_default_library",
+        "@com_google_cloud_go_storage//:go_default_library",
```
chaodaiG, line 14: move this one line lower

fejta: `bazel run @io_k8s_repo_infra//hack:update-bazel` will do this for you

pkg/merger/merger.go
chases2: I mean that the existing name, "shard", is good, and that you can delete the comment at the end.

pkg/merger/merger.go
chases2: Consider moving this `logrus.WithFields...` call outside the if statement instead. Writing the shards and their last-update-time to logs is still useful, even if dedicated metrics are disabled or missing.

"Mets is nil" is not as useful.

chases2 (commented):
Please squash your commits into one commit as well.

fejta:
https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git (I use `git rebase -i HEAD~6` and then change them from `pick` to `f` for fixup or `s` for squash if I want to keep the message, but there are 70 other options)

chases2 (approved):

PR #1013, opened 2022-06-13
Update to bazel 5.2.0

fejta:


chaodaiG:
/hold
if https://github.com/GoogleCloudPlatform/oss-test-infra/pull/1655 is merged ahead of https://github.com/GoogleCloudPlatform/testgrid/pull/1014, there is a chance that prow status reconciler might mark `test-testgrid-all` job succeed as the job is not defined anywhere

fejta:
I can't get this to pass so it isn't a concern 🙃

PR #1022, opened 2022-06-16
Remove --filter-columns flag

chases2:


fejta (approved):

fejta:
Looks like this breaks the tabulator bench tests

PR #1047, opened 2022-07-08
Add TimeStamp field to Result structure

alin-jurj:
The scope of PR is to add the TimeStamp field for Result structure. The JUnit XML format does not have the timestamp as an attribute.
This field will give us the start time for each test.
It will be used to group all the tests that faill with all the tests that run in the same time.


michelle192837:
Hi! Out of curiosity, do you have a more specific example of the use case this is enabling?

PR #838, opened 2022-03-10
junit: Use "message" attribute at Message func

qinqon:
At Ginkgo v2 the "message" attribute contains the
description from the assert, this change use that attribute to compose
the output of Message function.

PR #1055, opened 2022-08-16
fix: Add skipped status back for Ginkgo v2

chendave:
fix: https://github.com/kubernetes/kubernetes/issues/111510

Signed-off-by: Dave Chen <dave.jungler@gmail.com>

@BenTheElder @aojea @pohly

### DIFF
--- a/pkg/updater/gcs.go
+++ b/pkg/updater/gcs.go
@@ -277,7 +277,8 @@ func convertResult(log logrus.FieldLogger, nameCfg nameConfig, id string, header
 	// Append each result into the column
 	for _, suite := range result.suites {
 		for _, r := range flattenResults(suite.Suites.Suites...) {
-			if r.Skipped != nil && r.Skipped.Value == "" {
+			// "skipped" is the string that is always appended when the test is skipped without any reason in Ginkgo V2, e.g., "focus" is specified, and the test is skipped.
+			if r.Skipped != nil && r.Skipped.Value == "" && (r.Skipped.Message == "skipped" || r.Skipped.Message == "") {
 				continue
 			}
 			c := Cell{CellID: cellID}
--- a/pkg/updater/gcs_test.go
+++ b/pkg/updater/gcs_test.go
@@ -1828,8 +1828,67 @@ func TestConvertResult(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "Ginkgo V2 skipped with message",
+			nameCfg: nameConfig{
+				format: "%s",
+				parts:  []string{testsName},
+			},
+			result: gcsResult{
+				started: gcs.Started{
+					Started: metadata.Started{
+						Timestamp: now,
+					},
+				},
+				finished: gcs.Finished{
+					Finished: metadata.Finished{
+						Timestamp: pint(now + 1),
+						Passed:    &yes,
+					},
+				},
+				suites: []gcs.SuitesMeta{
+					{
+						Suites: &junit.Suites{
+							Suites: []junit.Suite{
+								{
+									Results: []junit.Result{
+										{
+											Name:    "visible skip non-default msg",
+											Skipped: &junit.Skipped{Message: *pstr("non-default message")},
+										},
+										{
+											Name:    "invisible skip default msg",
+											Skipped: &junit.Skipped{Message: *pstr("skipped")},
+										},
+										{
+											Name:    "invisible skip msg empty",
+											Skipped: &junit.Skipped{Message: *pstr("")},
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+			expected: InflatedColumn{
+				Column: &statepb.Column{
+					Started: float64(now * 1000),
+				},
+				Cells: map[string]Cell{
+					"." + overallRow: {
+						Result:  statuspb.TestStatus_PASS,
+						Metrics: setElapsed(nil, 1),
+					},
+					"visible skip non-default msg": {
+						Result:  statuspb.TestStatus_PASS_WITH_SKIPS,
+						Message: "non-default message",
+						Icon:    "S",
+					},
+				},
+			},
+		},
 	}
-
 	for _, tc := range cases {
 		t.Run(tc.name, func(t *testing.T) {
 			log := logrus.WithField("test name", tc.name)


aojea:
v3 or v2?

chendave:
> v3 or v2?

for v3, we read the status as well, no change needed for v2.

pkg/updater/gcs_test.go
```diff
@@ -1828,8 +1828,59 @@ func TestConvertResult(t *testing.T) {
...
+			},
+			expected: InflatedColumn{
+				Column: &statepb.Column{
+					Started: float64(now * 1000),
+				},
+				Cells: map[string]Cell{
+					"." + overallRow: {
+						Result:  statuspb.TestStatus_PASS,
+						Metrics: setElapsed(nil, 1),
+						//Icon:    "",
```
chendave, line 1874: remove this line

aojea:
> for v3, we read the status as well, no change needed for v2.

I'm lost, there is a ginkgo v3?

chendave:
lol, blame on amnesia.

pkg/updater/gcs.go
```diff
@@ -309,7 +309,7 @@ func convertResult(log logrus.FieldLogger, nameCfg nameConfig, id string, header
 				if c.Message != "" {
 					c.Icon = "F"
 				}
-			case r.Skipped != nil:
+			case r.Skipped != nil || r.Status == "skipped": // For compatiblity with Ginkgo V2, check the status as well.
```
aojea, line 312: seems to fail here

> ./pkg/updater/gcs.go:312:58: "compatiblity" is a misspelling of "compatibility"

```suggestion
			case r.Skipped != nil || r.Status == "skipped": // For compatibility with Ginkgo V2, check the status as well.
```

aojea:
I think that testgrid is correct, it seems the report for skipped entries is now empty https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/108095/pull-kubernetes-e2e-kind/1559743773649408000/artifacts/junit_01.xml

> <testcase name="[It] [sig-storage] In-tree Volumes [Driver: local][LocalVolumeType: tmpfs] [Testpattern: Dynamic PV (block volmode)] multiVolume [Slow] should access to two volumes with the same volume mode and retain data across pod recreation on different node" classname="Kubernetes e2e suite" status="skipped" time="0.000265002">
<skipped message="skipped - Driver local doesn't support DynamicPV -- skipping"/>
</testcase>

and before https://storage.googleapis.com/kubernetes-jenkins/logs/ci-kubernetes-e2e-gce-scale-correctness/1539216946044604416/artifacts/junit_01.xml it contained the message, now is part of the tag

> <testcase name="[sig-storage] In-tree Volumes [Driver: azure-disk] [Testpattern: Dynamic PV (default fs)] volumes should store data" classname="Kubernetes e2e suite" time="0.000936996">
<skipped>test/e2e/storage/framework/testsuite.go:51 Only supported for providers [azure] (not gce) test/e2e/storage/drivers/in_tree.go:1577</skipped>
</testcase>

aojea:
I think I've found the problem

chendave:
@aojea , this is why we need this change.

Ginkgo V2 updated the junit structure based on the new spec,
https://llg.cubic.org/docs/junit/

It was use to be a string value

In v1,
```
type JUnitSkipped struct {
	Message string `xml:",chardata"`
}
```

Now in v2, this is an `attr`, 

```
type JUnitSkipped struct {
	// Message maps onto "pending" if the test was marked pending, "skipped" if the test was marked skipped, and "skipped - REASON" if the user called Skip(REASON)
	Message string `xml:"message,attr"`
}
```

So, testgrid cannot get the value directly as it is now stored in the "Message" as a attribute.




Since the new junit spec is already supported in testgrid, pls see:
https://github.com/GoogleCloudPlatform/testgrid/pull/832
https://github.com/GoogleCloudPlatform/testgrid/pull/838


We just need to check the message attribute now for Ginkgo V2.

NOTE that "skipped" is always padded to `Message` attr,  this is corresponding to the case that `value` is empty is V1, so we need to filter this out, otherwise, the testcase should be run but was skipped for some reason, then they should show up in the testgrid.

aojea:
I think that is a ginkgo bug

pkg/updater/gcs.go
```diff
@@ -277,7 +277,8 @@ func convertResult(log logrus.FieldLogger, nameCfg nameConfig, id string, header
 	// Append each result into the column
 	for _, suite := range result.suites {
 		for _, r := range flattenResults(suite.Suites.Suites...) {
-			if r.Skipped != nil && r.Skipped.Value == "" {
+			// "skipped" is the string that is always appended when the test is skipped without any reason in Ginkgo V2, e.g., "focus" is specified, and the test is skipped.
+			if r.Skipped != nil && (r.Skipped.Value == "" && (r.Skipped.Message == "skipped" || r.Skipped.Message == "")) {
```
chendave, line 281: In v2:  `r.Skipped.Value` should be always empty.

In v1:  `r.Skipped.Message` should be always empty

aojea: yes

aojea: we do want to report the skipped ones, for ginkgo v2 we should not continue if r.Skipped != nil and r.Skipped.Message != ""

aojea: manage to repro now https://go.dev/play/p/wYK_PJ3QZtS?v=goprev , this is easier to understand

aojea: based on your finding Chen, this was the PR introducing the new format https://github.com/GoogleCloudPlatform/testgrid/pull/832/

that added a Value `chardata` , but my reading is that Skipped never has such field

https://llg.cubic.org/docs/junit/

```
      <!-- skipped can appear 0 or once. optional -->
      <skipped message=""   <!-- message/description string why the test case was skipped. optional -->
	  />

      <!-- error indicates that the test errored.
           An errored test had an unanticipated problem.
           For example an unchecked throwable (exception), crash or a problem with the implementation of the test.
           Contains as a text node relevant data for the error, for example a stack trace. optional -->
      <error message="" <!-- The error message. e.g., if a java exception is thrown, the return value of getMessage() -->
	     type=""    <!-- The type of error that occured. e.g., if a java execption is thrown the full class name of the exception. -->
	     >error description</error>

      <!-- failure indicates that the test failed.
           A failure is a condition which the code has explicitly failed by using the mechanisms for that purpose.
           For example via an assertEquals.
           Contains as a text node relevant data for the failure, e.g., a stack trace. optional -->
      <failure message="" <!-- The message specified in the assert. -->
	       type=""    <!-- The type of the assert. -->
	       >failure description</failure>

```

I think that we should fix the field removing the value and only continue if

```
	if r.Skipped != nil && r.Skipped.Message == "" {
				continue
			}
```

or always report the skipped entry

aojea: cc@ @michelle192837 @chizhg

chendave: > we do want to report the skipped ones, for ginkgo v2 we should not continue if r.Skipped != nil and r.Skipped.Message != ""

this won't be skipped, I think the testcases I added has already verified this.

chendave: > I think that we should fix the field removing the value

I don't think so, the `value` field is still needed for backward compatibility, iiuc, the `value` here is the unmarshalled from v1 `Message` field.

```
type JUnitSkipped struct {
	Message string `xml:",chardata"`
}
```

testgrid should support both the old and new format.

aojea: fair enough to me, I just said that because this moved from 

```
	Skipped    *string     `xml:"skipped,omitempty"`
```

to 

```
// Skipped holds <skipped/> elements.
type Skipped struct {
	Message string `xml:"message,attr"`
	Value   string `xml:",chardata"`
}
```

I let you decide, I don't have any strong opinion, just to keep the reports fully working again ;)

chendave: > or always report the skipped entry

k8s now has more than 7k+ e2e testcases, show all the skipped testcase with no useful information doesn't bring us something. 

thinking of the case that I run e2e and only focus on 1 testcase, we will only have only one valid result but all others are skipped, do we really all those skipped ones?

example of the skipped one in v2
```
          <testcase name="[It] [sig-storage] In-tree Volumes [Driver: azure-disk] [Testpattern: Pre-provisioned PV (default fs)] subPath should support file as subpath [LinuxOnly]" classname="Kubernetes e2e suite" status="skipped" time="0">
              <skipped message="skipped"></skipped>
          </testcase>
```


I think this is why testgrid doesn't show the skipped testcase which has the empty `value`.

```golang
if r.Skipped != nil && r.Skipped.Value == ""
```
and then treat it as invisible.

in v1,

```
<testcase name="[sig-storage] In-tree Volumes [Driver: nfs] [Testpattern: Dynamic PV (block volmode)] multiVolume [Slow] should access to two volumes with different volume mode and retain data across pod recreation on different node" classname="Kubernetes e2e suite" time="0">
<skipped></skipped>
</testcase>
```

I am new to this project, so all of those are just based on the assumption, so pls correct me if I miss something.

aojea: ah, no, you are right, that should be the reason then and an I fully agree with that, we should not report the empty ones

pohly: I agree with suppressing the "empty ones". But note that ginkgo v2 no longer skips them with an empty string. Instead it uses "skipped" as message. For example, when I run e2e.test with focus on a single test, I get for the others:
```
          <testcase name="[It] [sig-storage] In-tree Volumes [Driver: emptydir] [Testpattern: Dynamic PV (ntfs)][Feature:Windows] volumes should allow exec of files on the volume" classname="Kubernetes e2e suite" status="skipped" time="0">
              <skipped message="skipped"></skipped>
          </testcase>
```

We should also ignore those.

The interesting ones are the ones which were skipped at runtime. They tend to have a more useful explanation that IMHO is worthwhile showing in testgrid because a) running these tests cost resources and b) perhaps they were supposed to be running.

pohly: Never mind, it looks like the code already handles "message=skipped". Looks good to me.

michelle192837: (As an aside, thanks for clarifying why empty skipped cases aren't reported, I knew that was a thing and had forgotten why).

chizhg: This is a long thread and I haven't checked what it is all about, but there has been a Google-internal issue about TestGrid not showing skipped tests, and the bug number is 221515005

michelle192837, line 281: nit: Might be worth breaking this boolean into its own function and unit testing with a few cases.

I think you can also remove some of those parentheses.

chendave: - removed one extra parentheses
- add one more test to cover the case that msg is empty which should be invisible, for the test on the value it is already covered by those test: https://github.com/GoogleCloudPlatform/testgrid/blob/ec9e4dba3d782df007da1a90cdee0ab1916c749d/pkg/updater/gcs_test.go#L757-L764
- I am not sure we need to move this line to a function since it's just one line and I cannot see it could be reused somewhere, maybe we can do that when this check is needed in other places as well?

pkg/updater/gcs_test.go
```diff
@@ -1828,8 +1828,67 @@ func TestConvertResult(t *testing.T) {
...
+											Name:    "visible skip non-default msg",
+											Skipped: &junit.Skipped{Message: *pstr("non-default message")},
+										},
+										{
+											Name:    "invisible skip default msg",
+											Skipped: &junit.Skipped{Message: *pstr("skipped")},
+										},
+										{
+											Name:    "invisible skip msg empty",
+											Skipped: &junit.Skipped{Message: *pstr("")},
```
chendave, line 1865: this kind of should be tested for v1, we won't hit this situation in v2 afaik, add this to increase the test coverage.

pkg/updater/gcs.go
michelle192837: Doesn't have to be in a separate function, but even for 1-liners that aren't reused I've found a separate function is easier to test. Added coverage seems good though. ^^

michelle192837 (approved):
/hold

Looks good to me, holding in case other folks want to approve. Feel free to unhold at will.

aojea:
> /hold
> 
> Looks good to me, holding in case other folks want to approve. Feel free to unhold at will.

/hold cancel

:clap:

PR #1057, opened 2022-09-13
minor refactors to updater

chases2:
- Makes it so that the issues passed into the function don't have the same name as a temporary variable.
    - Rename also tries to make it clear what ll. 1031-1039 actually do: deduplicate the array
- Simplifies the SortStarted function signature; the config isn't useful


michelle192837 (approved):
Thanks Sean!

PR #1058, opened 2022-09-16
Implement tab state extension

chases2:
`--extend` on the Tabulator will now cause the tabulator to extend state with the information in the updater instead of simply overwriting it.

This feature is most useful for tabs with aggressive filtering: that freed up space can be used now to store more history.


michelle192837 (approved):
Awesome! :D

PR #1061, opened 2022-09-28
Added a feature to categorize tab as acceptably flaky.

sultan-duisenbay:
Notable changes:

Protos
- Added an option to configure maximum tolerable flakiness to config proto
- Added a flag to denote a tab as acceptably flaky to summary proto
- Added column-wise metrics to summary proto

Code
- Added logic to mark the tab as acceptably flaky &  unit tests for that
- Added validation logic to check if max_acceptable_flakiness is within permitted range (0.0 , 100.0)


chases2 (approved):
Overall this looks great, and I'm looking forward to testing it out. Experimental features should be behind a flag, though, so we can disable/enable them for different environments.

/hold

pkg/summarizer/summary.go
```diff
@@ -694,18 +694,23 @@ func updateTab(ctx context.Context, tab *configpb.DashboardTab, group *configpb.
 	alert := staleAlert(mod, latest, staleHours(tab))
 	failures := failingTestSummaries(grid.Rows)
 	passingCols, completedCols, passingCells, filledCells, brokenState := gridMetrics(len(grid.Columns), grid.Rows, recent, tab.BrokenColumnThreshold)
+	metrics := tabMetrics(passingCols, completedCols)
+	tabStatus := overallStatus(grid, recent, alert, brokenState, failures)
+	acceptablyFlaky := acceptableFlakiness(passingCols, completedCols, tabStatus, tab.GetStatusCustomizationOptions())
```
chases2, line 699: Ideally, we would add a flag to gate how statuses are calculated. I think you could do that here:
```
acceptablyFlaky := allowFuzzyFlakiness && acceptableFlakiness(...)
```

... then you could pass in the bool in updateTab (and in update, etc...) through the functions, or with a [package-level variable](https://github.com/GoogleCloudPlatform/testgrid/commit/86f7dd7a44839d93611f6cb4506d064e3d4acc48#diff-6f575e683c5fe1eb9301f314392e63c56a00e7b09ae9663adf4c3e4d67f81c70L786-L788).

See [this commit](https://github.com/GoogleCloudPlatform/testgrid/commit/86f7dd7a44839d93611f6cb4506d064e3d4acc48#diff-b2bac5c2bfa8982ae6b2f98b99cd59f6e8722857209295c2bc74baa83c7bc76fL117-L119) for an example (it's deleting a flag, but it shows what I mean by adding the flag)

sultan-duisenbay:
Makes sense. I will add a flag to cmd/summarizer/main and propagate it all the way down to that function.

pkg/summarizer/summary.go
sultan-duisenbay: Makes sense, I will add the flag and propagate it all the way down to this function. Thanks!

sultan-duisenbay: Done @chases2 & @michelle192837 ! I also added a couple of test cases to TestUpdateTab to see the overall behaviours of this newly added feature.

config/config.go
```diff
@@ -413,6 +413,11 @@ func validateDashboardTab(dt *configpb.DashboardTab) error {
 		}
 	}
 
+	// Max acceptable flakiness parameter should be valid.
+	if max_acceptable_flakiness := dt.GetStatusCustomizationOptions().GetMaxAcceptableFlakiness(); max_acceptable_flakiness < 0 || max_acceptable_flakiness > 100 {
+		mErr = multierror.Append(mErr, errors.New("invalid value provided for max_acceptable_flakiness"))
```
michelle192837, line 418: nit: Note what the valid value should be (e.g. "between 0.0 and 100.0, exclusive.")

config/config_test.go
```diff
@@ -990,6 +990,27 @@ func TestValidateDashboardTab(t *testing.T) {
 			},
 			err: true,
 		},
+		{
```
michelle192837, line 993: nit: check at the boundaries as well (0.0 and 100.0?)

michelle192837 (approved):
/hold

Looks good to me aside from nits, feel free to cancel the hold after addressing ^^
(though it might need re-approval, I forget).

config/config.go
sultan-duisenbay: Done

michelle192837: Ah sorry, I meant in the error message itself, so a user knows what to set the value to. ^^

michelle192837: Also I misread the logic condition, should be 'inclusive' not 'exclusive'.

michelle192837:
Thanks! `hold /cancel` at will ^^

michelle192837:
(Oh geez how did I typo that >< Thanks!)

PR #1070, opened 2022-10-31
Added a feature to ignore cols & enforce min # of runs

sultan-duisenbay:
#### Protos:
- **config**: Added config fields for ignorable test statuses & minimun runs required
- **summary**: Added a PENDING tab status, and ignored columns in metrics

#### Code
- **summarizer binary** - flags to opt-in features
- **status** - now accounts for min runs required & ignored columns
- **grid metrics** - returns ignored cols too
- **status message** - highlights tab stats & additional status info for selected tab statuses

- **nits** - aggregated gridMetrics and feature flags into structs

#### Tests
- individual features
- their combinations


pb/config/config.proto
```diff
@@ -579,7 +579,10 @@ message DashboardTab {
   DashboardTabFlakinessAlertOptions flakiness_alert_options = 24;
 
   // Configuration options for customizing dashboard tab status calculation.
-  // Used to define acceptable flakiness
+  // Options here define:
```
michelle192837, line 582: Think you can leave this off, and let the underlying message define the options.

pkg/summarizer/summary.go
```diff
@@ -668,7 +668,7 @@ func staleHours(tab *configpb.DashboardTab) time.Duration {
 }
 
 // updateTab reads the latest grid state for the tab and summarizes it.
-func updateTab(ctx context.Context, tab *configpb.DashboardTab, group *configpb.TestGroup, groupReader gridReader, allowFuzzyFlakiness bool) (*summarypb.DashboardTabSummary, error) {
+func updateTab(ctx context.Context, tab *configpb.DashboardTab, group *configpb.TestGroup, groupReader gridReader, allowFuzzyFlakiness, allowIgnoredColumns, allowMinNumberOfRuns bool) (*summarypb.DashboardTabSummary, error) {
```
michelle192837, line 671: I dunno that this is your responsibility, but since a lot of similar parameters/return values are getting passed around I think we might want to consider a struct to hold them, for less confusion down the road. (If you want to add a TODO to that effect, that would be awesome!)

The return from gridMetrics(), and passing all the boolean options around, are the main things I'm looking at.

sultan-duisenbay: that's a great idea! I went ahead and aggregated the feature flags and column/cell metrics into respective structs. LMK what you think. Thanks!

pkg/summarizer/summary.go
```diff
@@ -968,8 +980,17 @@ func allLinkedIssues(rows []*statepb.Row) []string {
 	return linkedIssues
 }
 
+// columnCellMetrics aggregates columnar and cellular metrics as a struct
+type columnCellMetrics struct {
```
michelle192837, line 984: nit: wondering if there's a better name for this and the associated variables? (gridMetrics seems great, but it's also the method name, and `metrics` is too generic/confusable with the module metrics. :( Maybe 'gridStats' or something?

@chases2 any suggestions here?

michelle192837 (commented):
Looks good overall, just the one nit from me!

pkg/summarizer/summary.go
sultan-duisenbay: I was thinking about gridMetrics too :) but didn't want to create confusion... `gridStats` actually sounds pretty good too.

sultan-duisenbay: changed the struct name to `gridStats` and variable name to metrics. PTAL

michelle192837 (approved):

PR #1092, opened 2022-12-01
Cache default config as map in API component

chases2:
Includes #1086; see second commit for review

Keeps a normalization cache for the API component only. For default configs, increases the normalization cost to constant instead of O(dashboard) or O(dashboard * tabs)

For non-default configs, front-loads the normalization needed without benefit. Keeping multiple cached configs in an expiring cache object in a later PR would improve repeated, non-default calls.

### DIFF
--- a/config/cache.go
+++ b/config/cache.go
@@ -30,6 +30,8 @@ import (
 	"github.com/GoogleCloudPlatform/testgrid/util/gcs"
 )
 
+// TODO(slchase): remove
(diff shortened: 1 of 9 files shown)


michelle192837 (approved):
Looks good to me!
/hold

Unhold at will

chases2:
Rebased and added a couple of info log lines; there's a theoretical case here where if the API is forever slammed, the config will never get around to updating, see [this example](https://go.dev/play/p/xzvUyZdcMkt).

I don't think this will happen, but the log line will let us confirm whether it is or not.

PR #1095, opened 2022-12-06
set ACCEPTABLE as a separate tab status

sultan-duisenbay:
Following our discussions, setting `ACCEPTABLE` as a distinct tab status in summary.proto


pb/summary/summary.proto
```diff
@@ -178,6 +178,7 @@ message DashboardTabSummary {
     STALE = 5;
     BROKEN = 6;
     PENDING = 7;
+    ACCEPTABLE = 8;
```
michelle192837, line 181: You can also mark the `acceptably_flaky` field as deprecated if you don't intend it to be used going forward. (similar to some other fields in this proto).

michelle192837 (approved):
LGTM

/hold for comment

Unhold at will!

michelle192837 (approved):

PR #1100, opened 2022-12-14
Initial Lit Component Demo

chases2:
Describes an application that uses the API to display a list of dashboards.

Currently not integrated with Bazel. Bazel integration (at least as far as the proto generation goes) will be added in a later CL. It's also not clear to me that this shouldn't live in a separate repository, but that would further complicate the proto import step.

/hold


michelle192837 (approved):
Squashing commits might be nice, but overall LGTM

chases2:
I'm looking into setting this up in a separate repository. It would be simpler to keep the TypeScript and Go code/build/test systems separate.

PR #1106, opened 2023-01-18
Remove Erik and add myself to reviewers/approvers

sultan-duisenbay:
I noticed I can't `lgtm` or `approve`, so this should help.


chases2:
Looks good overall, but I think we should add fejta to a new emeritus_approvers section. [Docs](https://github.com/kubernetes/community/blob/master/contributors/guide/owners.md#emeritus) and [example](https://github.com/kubernetes/test-infra/blob/0dc10ffb097f889b57796f77fc6c8d70d4bb8dd5/OWNERS#L18) from the k8s community here.

sultan-duisenbay:
Added `fejta` to `emeritus_approvers`.

sultan-duisenbay:
> Looks good overall, but I think we should add fejta to a new emeritus_approvers section. [Docs](https://github.com/kubernetes/community/blob/master/contributors/guide/owners.md#emeritus) and [example](https://github.com/kubernetes/test-infra/blob/0dc10ffb097f889b57796f77fc6c8d70d4bb8dd5/OWNERS#L18) from the k8s community here.

Added fejta to `emeritus_approvers`. Thanks!

chases2 (approved):

PR #1118, opened 2023-02-03
Implement ignore_skip to exclude skipped results.

michelle192837:
This is flag-guarded so the behavior will not kick in until the flag is set on deployment. Adding this so ignored results properly get excluded when specified in config.

### DIFF
--- a/cmd/updater/main.go
+++ b/cmd/updater/main.go
@@ -51,6 +51,7 @@ type options struct {
 	gridPrefix       string
 	subscriptions    util.Strings
 	reprocessList    util.Strings
+	enableIgnoreSkip bool
 
 	debug    bool
 	trace    bool
@@ -110,6 +111,8 @@ func gatherFlagOptions(fs *flag.FlagSet, args ...string) options {
 	fs.DurationVar(&o.buildTimeout, "build-timeout", 3*time.Minute, "Maximum time to wait to read each build")
 	fs.StringVar(&o.gridPrefix, "grid-prefix", "grid", "Join this with the grid name to create the GCS suffix")
 
+	fs.BoolVar(&o.enableIgnoreSkip, "enable-ignore-skip", false, "If true, enable ignore_skip behavior.")
+
 	fs.BoolVar(&o.debug, "debug", false, "Log debug lines if set")
 	fs.BoolVar(&o.trace, "trace", false, "Log trace and debug lines if set")
 	fs.BoolVar(&o.jsonLogs, "json-logs", false, "Uses a json logrus formatter when set")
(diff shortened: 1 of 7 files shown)


cmd/updater/main.go
```diff
@@ -160,7 +163,7 @@ func main() {
 	})
 	log.Info("Configured concurrency")
 
-	groupUpdater := updater.GCS(ctx, client, opt.groupTimeout, opt.buildTimeout, opt.buildConcurrency, opt.confirm)
+	groupUpdater := updater.GCS(ctx, client, opt.groupTimeout, opt.buildTimeout, opt.buildConcurrency, opt.confirm, opt.enableIgnoreSkip)
```
sultan-duisenbay, line 166: not required for this PR. I see a lot of crucial methods are being called with increasing number of parameters, maybe we should aggregating those and pass the aggregated structs of params

sultan-duisenbay:
Some test cases complain about not having enough args in them.
- pkg/updater/read_test.go:1018:103: not enough arguments in call to readColumns
- pkg/updater/updater_test.go:98:41: not enough arguments in call to GCS
- pkg/updater/updater_test.go:428:101: not enough arguments in call to GCS
- pkg/updater/updater_test.go:2162:59: not enough arguments in call to gcsColumnReader

cmd/updater/main.go
michelle192837: Agreed! Filed https://github.com/GoogleCloudPlatform/testgrid/issues/1121 to track ^^

pkg/updater/gcs_test.go
```diff
@@ -274,7 +274,7 @@ func TestSplitCells(t *testing.T) {
 			},
 		},
 		{
-			name: "many items eventually truncate",
+			name: "minterface{} items eventually truncate",
```
sultan-duisenbay, line 277: those pesky "replace all"s are elusive :)

pkg/updater/gcs_test.go
```diff
@@ -890,7 +890,7 @@ func TestConvertResult(t *testing.T) {
 											},
 										},
 										{
-											Name: "many digits",
+											Name: "minterface{} digits",
```
sultan-duisenbay, line 893: ditto

pkg/updater/gcs_test.go
```diff
@@ -965,7 +965,7 @@ func TestConvertResult(t *testing.T) {
 							"food": 123456789,
 						},
 					},
-					"many digits": {
+					"minterface{} digits": {
```
sultan-duisenbay, line 968: ditto

sultan-duisenbay (commented):
Looks good to me, there are some instance were `many` were replaced to `minstance` - once this is done should be good 2 go!

pkg/updater/gcs_test.go
michelle192837: Whoops 😬  Good catch, thanks!

sultan-duisenbay (approved):
LGTM!

Issue #1124, opened 2023-02-09
Add nanos to timestamp when converting


michelle192837:
https://github.com/GoogleCloudPlatform/testgrid/blob/4d8396ffaf3e1df48f53a374ca7989af30678ce2/pkg/api/v1/summary.go#L181

We convert a double-typed timestamp to a proto timestamp (which has 'seconds' and 'nanos' fields), but we drop nanoseconds when doing so. Instead, add them when converting.

ankur12-1610:
I would like to work on this issue!

gunwant11:
@michelle192837  I can pick up this issue if its not assigned :)

michelle192837:
Going to let first-come-first-serve apply here, so Ankur feel free to assign to yourself ^^ (I think `/assign` should work)

ankur12-1610:
> Going to let first-come-first-serve apply here, so Ankur feel free to assign to yourself ^^ (I think `/assign` should work)

Got it!

PR #1131, opened 2023-02-16
Added nanos to timestamp when converting

ankur12-1610:
Added `Nanos` to the `timestamp` in `convertSummary` function
Fixes: #1124

### DIFF
--- a/pkg/api/v1/summary.go
+++ b/pkg/api/v1/summary.go
@@ -19,6 +19,7 @@ package v1
 import (
 	"context"
 	"fmt"
+	"math"
 	"net/http"
 
 	"github.com/gorilla/mux"
@@ -47,16 +48,20 @@ var (
 
 // convertSummary converts the tab summary from storage format (summary.proto) to wire format (data.proto)
 func convertSummary(tabSummary *summarypb.DashboardTabSummary) *apipb.TabSummary {
+	RSeconds, RNanos := math.Modf(tabSummary.LastRunTimestamp)
+	USeconds, UNanos := math.Modf(tabSummary.LastUpdateTimestamp)
 	return &apipb.TabSummary{
 		DashboardName:         tabSummary.DashboardName,
 		TabName:               tabSummary.DashboardTabName,
 		OverallStatus:         tabStatusStr[tabSummary.OverallStatus],
 		DetailedStatusMessage: tabSummary.Status,
 		LastRunTimestamp: &timestamp.Timestamp{
-			Seconds: int64(tabSummary.LastRunTimestamp),
+			Seconds: int64(RSeconds),
+			Nanos:   int32(RNanos * 1e9),
 		},
 		LastUpdateTimestamp: &timestamp.Timestamp{
-			Seconds: int64(tabSummary.LastUpdateTimestamp),
+			Seconds: int64(USeconds),
+			Nanos:   int32(UNanos * 1e9),
 		},
 		LatestPassingBuild: tabSummary.LatestGreen,
 	}
--- a/pkg/api/v1/summary.go
+++ b/pkg/api/v1/summary.go
@@ -48,20 +48,20 @@ var (
 
 // convertSummary converts the tab summary from storage format (summary.proto) to wire format (data.proto)
 func convertSummary(tabSummary *summarypb.DashboardTabSummary) *apipb.TabSummary {
-	RSeconds, RNanos := math.Modf(tabSummary.LastRunTimestamp)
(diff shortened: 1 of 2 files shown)


ankur12-1610:
@michelle192837 can you please review?

pkg/api/v1/summary.go
```diff
@@ -54,9 +54,11 @@ func convertSummary(tabSummary *summarypb.DashboardTabSummary) *apipb.TabSummary
 		DetailedStatusMessage: tabSummary.Status,
 		LastRunTimestamp: &timestamp.Timestamp{
 			Seconds: int64(tabSummary.LastRunTimestamp),
+			Nanos:   int32(tabSummary.LastRunTimestamp),
```
sultan-duisenbay, line 57: @ankur12-1610 the idea is to convert the fractional part of the `LastRunTimestamp`. We are currently just rounding it off to fetch seconds but discarding the fractional part.

ankur12-1610: > @ankur12-1610 the idea is to convert the fractional part of the `LastRunTimestamp` coming from `tabSummary` (it is a double there - hence might contain some fractional part of timestamp). We are currently just rounding it off to fetch seconds but discarding the fractional part.
> 
> The fractional part might be useful and needs to be added in nanos to timestamp struct. Once done, the test cases need to be updated too to verify the correctness. Thanks!

Ah! got it!
I've updated the code and used `modf()` for that, I haven't written any tests will do so once the current implemented method is confirmed. Can you please verify?

sultan-duisenbay (commented):
LGTM! Please add the test cases now :) - thanks!

pkg/api/v1/summary.go
sultan-duisenbay: looks good to me @ankur12-1610 ! thanks!

ankur12-1610:
> LGTM! Please add the test cases now :) - thanks!

Hey, I've added tests, should I create a different test for Nanos or I should insert it in existing tests as I've done in the recent commit?

sultan-duisenbay (commented):
Would be curios to see additional test cases with a variable number of digits in the fractional part (.354 or even .1). Could you please add those too @ankur12-1610 ?

pkg/api/v1/summary_test.go
```diff
@@ -500,9 +508,11 @@ func TestListTabSummariesHTTP(t *testing.T) {
 						LatestPassingBuild:    "Lantern",
 						LastUpdateTimestamp: &timestamp.Timestamp{
 							Seconds: 916166800,
+							Nanos:   0,
```
sultan-duisenbay, line 511: the default value for number fields in Go is 0, so you don't have specify in that case

ankur12-1610: Got it!

ankur12-1610:
> Would be curios to see additional test cases with a variable number of digits in the fractional part (.354 or even .1). Could you please add those too @ankur12-1610 ?

@sultan-duisenbay should these be enough?

sultan-duisenbay (changes requested):
Sorry for the long turnaround time @ankur12-1610 . Today is a day off in the US and Canada.

I believe you might have missed a typo in one of your test cases. I would advise to run the tests locally first (much faster that way) using `blaze test //...` command

pkg/api/v1/summary_test.go
```diff
@@ -103,18 +111,36 @@ func TestListTabSummaries(t *testing.T) {
...
 							LastRunTimestamp:    float64(916166800),
 						},
+						{
+							DashboardName:       "Marco",
+							DashboardTabName:    "polo-4",
+							Status:              "1/7 tests are failing!",
+							OverallStatus:       summarypb.DashboardTabSummary_ACCEPTABLE,
+							LatestGreen:         "Lantern",
+							LastUpdateTimestamp: float64(0.916166782),
+							LastRunTimestamp:   915166800.9161667820.916166782),
```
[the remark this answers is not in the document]
ankur12-1610: Ah! I forgot to update the code :sweat_smile:
Ran `bazel test //...` and tests related to this file were all **passing**

ankur12-1610:
> Sorry for the long turnaround time @ankur12-1610 . Today is a day off in the US and Canada.
> 
> I believe you might have missed a typo in one of your test cases. I would advise to run the tests locally first (much faster that way) using `blaze test //...` command

Np :smile: 
Ah! I forgot to update the code :sweat_smile:
Ran `bazel test //...` and tests related to this file were all **passing**

sultan-duisenbay (approved):
LGTM!

Issue #1123, opened 2023-02-09
Remove use of 'grid path prefix' where it's no longer necessary


michelle192837:
Updates used to go through the Updater (grab results, make a grid) to the Summarizer (summarize stats for a grid).

Now there's also Tabulator, so updates go from the Updater > Tabulator (turn a grid into the user-viewed grid) > Summarizer.

We still have some leftover flags or code though that are there from the migration to using the Tabulator module. For example:
- https://github.com/GoogleCloudPlatform/testgrid/blob/f325852c59aa08e560990644f772c7c04432c6ef/pkg/api/v1/state.go#L85 and https://github.com/GoogleCloudPlatform/testgrid/blob/f325852c59aa08e560990644f772c7c04432c6ef/pkg/api/v1/server.go#L32 (GridPrefixPath should never be set, we should be able to remove that setting and any code that uses it)
- https://github.com/GoogleCloudPlatform/testgrid/blob/18fbd6a6712adb48e2427daaa5b07752a0a63cc3/cmd/summarizer/main.go#L45 (same here, no need for gridPathPrefix)

ankur12-1610:
@michelle192837 I would like to work on this as the current issue #1124 which I'm working on is almost solved.

michelle192837:
Sounds good! Feel free to assign to yourself ^^

ankur12-1610:
Sure
/assign

PR #1138, opened 2023-02-23
Updated path prefixes

ankur12-1610:
Updated the path prefixes according to the new flow: `Updater > Tabulator (turn a grid into the user-viewed grid) > Summarizer`

Fixes: #1123

### DIFF
--- a/pkg/api/v1/state.go
+++ b/pkg/api/v1/state.go
@@ -65,7 +65,7 @@ func findDashboardTab(cfg *cachedConfig, dashboardInput string, tabInput string)
 
 // GroupGrid fetch tab group name grid info (columns, rows, ..etc)
 func (s Server) GroupGrid(ctx context.Context, configPath *gcs.Path, groupName string) (*statepb.Grid, error) {
-	groupPath, err := configPath.ResolveReference(&url.URL{Path: path.Join(s.GridPathPrefix, groupName)})
+	groupPath, err := configPath.ResolveReference(&url.URL{Path: path.Join(s.TabPathPrefix, groupName)})
 	if err != nil {
 		return nil, fmt.Errorf("resolve: %v", err)
 	}
@@ -81,9 +81,6 @@ func (s Server) Grid(ctx context.Context, scope string, dashboardName, tabName, 
 	configPath, _, err := s.configPath(scope)
 	if err != nil {
 		return nil, err
-	}
-	if s.TabPathPrefix == "" { // TODO(chases2): Delete; all APIs should be configured to use Tabulator now
-		return s.GroupGrid(ctx, configPath, testGroupNanme)
 	}
 	path, err := tabulator.TabStatePath(*configPath, s.TabPathPrefix, dashboardName, tabName)
 	if err != nil {
--- a/pkg/api/v1/state.go
+++ b/pkg/api/v1/state.go
@@ -22,8 +22,6 @@ import (
 	"fmt"
 	"math"
 	"net/http"
-	"net/url"
-	"path"
 
 	"github.com/golang/protobuf/ptypes/timestamp"
 	"github.com/gorilla/mux"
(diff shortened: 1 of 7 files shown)


pkg/api/v1/state.go
```diff
@@ -65,7 +65,7 @@ func findDashboardTab(cfg *cachedConfig, dashboardInput string, tabInput string)
 
 // GroupGrid fetch tab group name grid info (columns, rows, ..etc)
 func (s Server) GroupGrid(ctx context.Context, configPath *gcs.Path, groupName string) (*statepb.Grid, error) {
```
michelle192837, line 67: nit: I think GroupGrid() shouldn't be referenced anymore with the lines below removed, so it should be safe to delete entirely. (It's good to confirm that's the case though).

michelle192837 (changes requested):
Think this looks good, thanks! Left a comment.

In addition, last action before this is good to approve:
- Manually verify that none of the removed flags are referenced in files under https://github.com/GoogleCloudPlatform/testgrid/tree/master/cluster

pkg/api/v1/state.go
sultan-duisenbay: +1 to this. This fetches the data from the updater rather than tabulator

sultan-duisenbay (changes requested):
@ankur12-1610 this looks good to me. However, the tests are failing and I think I know why. since we removed the gridpathprefix, we are not looking in the grid path folder anymore and defaulting to tabs path prefix.

change the `serverTabPathPrefix` in the fake to "tabs"
in all the API test cases where the Grid is used, change the path to `gs://{bucket}/tabs/{dashboard_name}/{tab_name}` from `gs://{bucket}/grid/{testgroup_name}` and see if this helps

ankur12-1610:
@michelle192837 @sultan-duisenbay fixed the tests, they are green now and have updated the code according to the reviews :)

Verified https://github.com/GoogleCloudPlatform/testgrid/tree/master/cluster as well, nothing to change.

ankur12-1610:
Screenshot from 2023-02-25 12-46-41

These are the files that still have `GridPrefixPath`, I didn't remove because I thought it was related to the tabulator which requires `GridPrefixPath`.

chases2:
> These are the files that still have `GridPrefixPath`, I didn't remove it because I thought it was related to the tabulator which requires `GridPrefixPath`.

That makes sense. The tabulator should have a configurable "grid" prefix (to read from) and "tabs" prefix (to write to).

Another way to think of it is that the "grid" is the output of the updater (to grab data) and "tabs" is the output of the tabulator (to refine data). They output the same ["shape"](https://github.com/GoogleCloudPlatform/testgrid/blob/master/pb/state/state.proto) of data, just different amounts of it.

chases2:
Looks good to me, but I'll leave it to @michelle192837 (who filed the inital bug) to approve.

/lgtm
/lint

ankur12-1610:
> > These are the files that still have `GridPrefixPath`, I didn't remove it because I thought it was related to the tabulator which requires `GridPrefixPath`.
> 
> That makes sense. The tabulator should have a configurable "grid" prefix (to read from) and "tabs" prefix (to write to).
> 
> Another way to think of it is that the "grid" is the output of the updater (to grab data) and "tabs" is the output of the tabulator (to refine data). They output the same ["shape"](https://github.com/GoogleCloudPlatform/testgrid/blob/master/pb/state/state.proto) of data, just different amounts of it.

Yeah got it. At the start, it was a little bit confusing about what to ignore and what not to, so to keep it simple I just ignored all the Tabulator-related occurrences :sweat_smile:

sultan-duisenbay:
> @michelle192837 @sultan-duisenbay fixed the tests, they are green now and have updated the code according to the reviews :)
> 
> Verified https://github.com/GoogleCloudPlatform/testgrid/tree/master/cluster as well, nothing to change.

thanks @ankur12-1610 - LGTM!

sultan-duisenbay (approved):

Issue #1120, opened 2023-02-07
[Needs improvement]Documentation for Install and use.


RitheshShenthar:
Currently, the documentation for Testgrid is very high level and does not have an "Installation and usage" section similar to kind's https://kind.sigs.k8s.io/.
The info provided lacks the implementation details to get a quick start with open source tools to setup the Testgrid backend. More specifically, we need the exact commands to do the following:

How to Install, build and run a testgrid instance running on localhost working with a local storage server.
K8S config file for deploying the OSS parts of testgrid as an app which can be integrated by a FE service.
Nice to have: An example of how we can customize a front end to consume the OSS Testgrid backend.

michelle192837:
Updating a little bit for now. A proper update is going to take longer than I thought given other things we're working on (apologies), and I'd like to make sure instructions will be actually helpful and easy to use, so going to chip away at this and try to have things in a much more friendly state before mid-March. 

In the meanwhile I can clarify a few of these points:

1. Using this as a base to refine into a more useable quick start:
- `mkdir /tmp/testgrid`
- Create a configuration (config.proto), save at `/tmp/testgrid/config`
- Run `bazel run //cmd/updater:updater -- --config=/tmp/testgrid/config`
- Verify that results updated locally (`ls /tmp/testgrid/grid`).
Though note that a complete update is "test results" > Updater > Tabulator > Summarizer, stored locally if you want to verify protos programmatically with a script (ex: https://github.com/GoogleCloudPlatform/testgrid/tree/master/cmd/state_comparer), or in GCS to view on the frontend.

2. Yup, this would be nice to have. In the meanwhile, https://github.com/kubernetes/test-infra/tree/master/gubernator#gcs-layout has the expected format for results in GCS.
3. At the moment you'll need GCS or local files. (For using the frontend as a service, e.g. https://github.com/GoogleCloudPlatform/testgrid/blob/master/standalone.md#frontend-usage, that definitely requires GCS. For running some components locally like in (1), you can read/write to the local filesystem. I haven't confirmed, but I _think_ reading test results should work with local files as well, the example I ran locally just works with an existing config that uses GCS).

And for deployment, it's a bit hidden, but https://github.com/GoogleCloudPlatform/testgrid/blob/master/standalone.md#load-data has a link to an example deployment that other deployments can base off of.

TODOs (roughly this order):
[ ] Update standalone.md to clarify a few things and update information.
[...]

PR #1139, opened 2023-02-23
added metrics to keep count of "unread columns" in updater component.

### LINKED ISSUE #1101: Track a 'more' metric for the Updater
The Updater logs when it has more work to do after updating a test group (with 'more = true' in the log line), but it'd be nice to have a metric that allows us to track if there are spikes in groups that don't update, or a consistent line of groups that can't catch up on repeated updates. (This is expected in small numbers because some group updates, especially for new groups, take a long time, but it's a bad sign if this climbs or stays high).

moki1202:
Fixes #1101


moki1202:
@michelle192837 tried my best here. There are errors for these changes that need fixing but am I headed in the right direction? Also the names that I've decided for the metric field might just not be right :sweat_smile:

michelle192837:
> @michelle192837 tried my best here. There are errors for these changes that need fixing but am I headed in the right direction? Also the names that I've decided for the metric field might just not be right 😅

I think this looks good! I agree another name might be a bit clearer ('IncompleteUpdates' or something similar, for instance?).

moki1202:
> > @michelle192837 tried my best here. There are errors for these changes that need fixing but am I headed in the right direction? Also the names that I've decided for the metric field might just not be right 😅
> 
> I think this looks good so far. ^^ I agree another name might be a bit clearer ('IncompleteUpdates' or something similar, for instance?).

@michelle192837 done!

pkg/updater/updater.go
```diff
@@ -107,7 +108,8 @@ func GCS(poolCtx context.Context, colClient gcs.Client, groupTimeout, buildTimeo
 		defer cancel()
 		gcsColReader := gcsColumnReader(colClient, buildTimeout, readResult, enableIgnoreSkip)
 		reprocess := 20 * time.Minute // allow 20m for prow to finish uploading artifacts
-		return InflateDropAppend(ctx, log, client, tg, gridPath, write, gcsColReader, reprocess)
+		mets := CreateMetrics(prometheus.NewFactory())
```
michelle192837, line 111: Ah! So actually, this doesn't need to be created here, the metrics get created in the corresponding 'main.go' file under cmd/, e.g. https://github.com/GoogleCloudPlatform/testgrid/blob/master/cmd/updater/main.go#L168. (I believe you're already good to go and can revert these two lines.)

moki1202: @michelle192837 Understood! just 1 more doubt. Here, we pass `mets` as a param to `InflateDropAppend`, so the return statement needs that `mets` param. What should I pass here?

chases2: Bubbling it down from update.Update(), refactoring functions along the way, is probably the way to go. So, in this case, adding it to `func GCS(...)` would be my approach.

michelle192837: Ah sorry, I missed adding this to InflateDropAppend's call somehow. +1 to what Sean said!

moki1202: @chases2 I don't understand this. 😅

chases2: Try changing its signature to "func GCS(poolCtx context.Context, colClient gcs.Client, **`mets *Metrics`**, groupTimeout, buildTimeout time.Duration, concurrency int, write bool, enableIgnoreSkip bool) GroupUpdater"

moki1202: Oh I get it now!

moki1202:
@michelle192837 @chases2 does this look good now?

pkg/updater/updater.go
```diff
@@ -49,15 +49,17 @@ const componentName = "updater"
...
 }
 
 // CreateMetrics creates metrics for this controller
 func CreateMetrics(factory metrics.Factory) *Metrics {
 	return &Metrics{
-		UpdateState:  factory.NewCyclic(componentName),
-		DelaySeconds: factory.NewDuration("delay", "Seconds updater is behind schedule", "component"),
+		UpdateState:       factory.NewCyclic(componentName),
+		DelaySeconds:      factory.NewDuration("delay", "Seconds updater is behind schedule", "component"),
+		IncompleteUpdates: factory.NewCounter("counter", "number of unread columns"),
```
chases2, line 62: Please name this counter something more descriptive, such as "incomplete-updates".
Amend the description also, please. This counts the number of update attempts that don't complete, not the number of columns that were skipped over.

Note [here](https://github.com/GoogleCloudPlatform/testgrid/blob/f7fef68af4595f39ca416c9e3a070bc8885a5165/util/metrics/metrics.go#L48) that the arguments for NewCounter are (name, description, ...any additional fields to capture)

chases2:
We need a bit more documentation on this (#1120), but you can also check the way your new metric will look in prometheus if you'd like:

1. Run the updater with `bazel run //cmd/updater -- --wait=1h`
    - The "[--wait](https://github.com/GoogleCloudPlatform/testgrid/blob/f7fef68af4595f39ca416c9e3a070bc8885a5165/cmd/updater/main.go#L106)" flag makes it rerun continuously 
2. Go to localhost:2112, [the metrics port for prometheus](https://github.com/GoogleCloudPlatform/testgrid/blob/f7fef68af4595f39ca416c9e3a070bc8885a5165/util/metrics/prometheus/prometheus.go#L44)

chases2:
@moki1202 Are you interested in working on this still? I think this would be a serious improvement, and it seems like it's most of the way complete with what you have currently.

moki1202:
Nooooo! I totally forgot about this! @chases2 Thanks for the reminder. Will definitely complete this ASAP.

moki1202:
@chases2 I've pushed the remaining changes. Please let me know where this needs improvement. I'll try to fix it asap.

michelle192837:
Changes look good to me! Only issue is I'm not sure why the unit test is failing, going to retest one more time but if it's consistent it might need another look to see if something from this PR is affecting it.

michelle192837:
You can also check if this is consistent with your change, `bazel test //pkg/... --runs_per_test=100`.
That is weirdly consistent for no reason I can tell, but it's not failing like this at HEAD. Might be worth rebasing if you haven't already?

moki1202:
@michelle192837 I've rebased my branch. The unit tests still fail for some reason.

michelle192837:
> From these logs: https://oss.gprow.dev/view/gs/oss-prow/pr-logs/pull/GoogleCloudPlatform_testgrid/1139/test-testgrid-all/1663927888350547968#1:build-log.txt%3A189-190
> 
> Something is going on with these code changes that is causing the update function to upload to files... a different number of times? It's not clear to me why that's important for this test. I do wonder if it's flaky or exactly what's going on, but I don't think this should be blocking.
> 
> @michelle192837 Has this updater test flaked before? I feel like I remember something like that in the past.
> 
> Consider modifying the test to expect the new generation for now.

I think it has flaked before, though not in this specific way (there's a couple other cases that I think are flaky, but they aren't replicating for me with 100 runs from head).

moki1202:
@michelle192837 @chases2 is there something that I can do to help here?

PR #1141, opened 2023-02-24
Update README files under cmd/ with common tips and commands.

michelle192837:
I gave these a basic pass that these will run, though if you spot any issues please let me know.

Also cleaned up some old docs from when we were migrating, and tried to make these more consistent (with common tips, etc. in the cmd/ README instead of individual component READMEs).

Ref #1120


sultan-duisenbay (approved):
LGTM!

Issue #1116, opened 2023-02-01
Grid showing recent random builds (?) with Unknown: Build too old to process


drewhagen:
In CI Builds for Kubernetes SIG Release on test grid, we are seeing (?) pop up for "overall". When hovered over, the message is `Unknown: Build is too old to process @ <build-id> (<start time>)`

Kubernetes SIG Testing has dug into the code a little bit in a [Slack thread here](https://kubernetes.slack.com/archives/C09QZ4DQB/p1675193408560939?thread_ts=1675117747.178039&cid=C09QZ4DQB).
It seems this is expected behavior for builds that are more than 15 days old, but this is not the case for us.

### Replicating
1. Go to this test grid here:
https://k8s-testgrid.appspot.com/sig-release-master-blocking#gce-cos-master-scalability-100&show-stale-tests=
(making sure that show-stale-tests is shown and "Overall" can be seen)
2. Scroll horizontally over until a test has blank cells in it's column. A "?" should be in "Overall" with `Unknown: Build is too old to process @ <build-id> (<start time>)`

Build number 1620329353227276288 at 1/31 1:01 CST and build number 1620220887515009024 at 1/30 17:51 CST are showing this, even though it is 1/31.
![image](https://user-images.githubusercontent.com/2896571/215919805-9d12db13-e023-40b7-b4e2-80ced8bcb69b.png)

In the summary page, these (?) builds are counted as "flaky":
![image](https://user-images.githubusercontent.com/2896571/215920368-66456ed1-88a2-4969-a6c2-7ce0a71fd2f8.png)
![image](https://user-images.githubusercontent.com/2896571/215920265-4eb4fe88-ac57-4618-985d-9d27d8001ada.png)

michelle192837:
Another recent example of this: https://testgrid.k8s.io/google-gce#gce-cos-master-scalability-100 (https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gci-gce-scalability/1625083784154058752 for instance)

This is supposed to warn when columns can't be read because they're too old (and we have to stop processing at some point), but agree this is too recent to be reasonable. (Probably happening around https://github.com/GoogleCloudPlatform/testgrid/blob/4d8396ffaf3e1df48f53a374ca7989af30678ce2/pkg/updater/read.go#L188, the column is before the 'stop' time we list for some reason? Possibly there's interleaved builds, or something about the start time is not working as expected?).

[...]

PR #1143, opened 2023-02-24
Log more on 'too old' cols when processing started.json

michelle192837:
Trying to log more info when we hit https://github.com/GoogleCloudPlatform/testgrid/issues/1116. It's unclear what times the Updater is dealing with, so make it clear exactly what it's comparing.


sultan-duisenbay (approved):
LGTM!

PR #1147, opened 2023-02-28
Move API from Gorilla to Chi

chases2:
- Gorilla Mux is [archived](https://github.com/gorilla#gorilla-toolkit)
- The only features we're using are URL variables and subrouters, so a lighter-weight library is good, but using the http.ServeMux built-in isn't _quite_ enough.


michelle192837 (approved):
Looks good, thanks! ^^

sultan-duisenbay (approved):
LGTM!

PR #1155, opened 2023-03-04
Allow custom target status based on targetResult

XuanWang-Amos:
### Description

We have lots of tool failures in our tests which we want to override status to `abort`, existing `CustomStatus` function only works for `testResult`, but for tool failures, we won't have any testResult since we never run any test case.

This PR adds a `CustomTargetStatus` function which allows comparison and override based on `targetResult` (Currently it only allows status equal comparison).

### Testing
- Passed all existing bazel `go_default_test`.
- Also added test cases for newly added `CustomTargetStatus` function.


pkg/updater/eval.go
```diff
@@ -231,3 +253,37 @@ func evalProperties(rule *evalpb.Rule, testResult TestResult) *tspb.TestStatus {
...
 }
+
+func evalTargetProperties(rule *evalpb.Rule, targetResult TargetResult) *tspb.TestStatus {
+	for _, cmp := range rule.TestResultComparisons {
+		if cmp.Comparison == nil {
+			return nil
+		}
+		var stscmp func(tspb.TestStatus, tspb.TestStatus) bool
+		stsval := cmp.Comparison.GetTargetStatusValue()
+		switch cmp.Comparison.Op {
```
chases2, line 264: If you don't plan immediately to implement other comparison operations (and I think equals alone is fine), make this a simple if statement that returns early.

```
// Only EQ is supported
if cmp.Comparision.Op != evalpb.Comparision_OP_EQ {
    return nil
}
```

pkg/updater/eval.go
```diff
@@ -231,3 +253,37 @@ func evalProperties(rule *evalpb.Rule, testResult TestResult) *tspb.TestStatus {
...
+		case evalpb.Comparison_OP_EQ:
+			stscmp = targetStatusEQ
+		}
+		if f := cmp.GetTargetResultField(); f != "" {
+			var getSts func() tspb.TestStatus
+			switch f {
+			case "target_status":
```
chases2, line 271: Same here, although I think the if statement could be rolled into line 268

```
// target statuses only
if f := cmp.GetTargetResultField(); f ==  "target_status" {
   ...
}
```

pkg/updater/eval_test.go
```diff
@@ -371,3 +371,93 @@ func TestCustomStatus(t *testing.T) {
...
+								Op: evalpb.Comparison_OP_EQ,
+								ComparisonValue: &evalpb.Comparison_TargetStatusValue{
+									TargetStatusValue: tspb.TestStatus_CATEGORIZED_ABORT,
+								},
+							},
+						},
+					},
+				},
+			},
+			tgr: makeTargetResult(tspb.TestStatus_TOOL_FAIL),
```
chases2, line 444: nit: this test would be clearer to me (in comparison to the last test) if you added `want: nil` here.

pb/custom_evaluator/custom_evaluator.proto
```diff
@@ -52,6 +52,12 @@ message TestResultComparison {
     //
     // NOTE: Only supported for string and numerical values
     string test_result_error_field = 4;
+
+    // This will find the scalar field with the given name within the TargetResult
+    // proto. The value of that field will be used to evaluate.
+    //
+    // NOTE: Only supported for TestStatus values.
+    string target_result_field = 5;
```
chases2, line 60: Is this the best way to configure a feature like this? Why not a boolean for "target_status"? What else would a user want to set it to?

chases2 (commented):
Thanks for the contribution! I have a couple of comments on the code, mostly around switch statements.

I'd also like to kick this over to Michelle for a review of the Custom Evaluator proto change (since that's not my area of expertise).

/assign @michelle192837 
/ok-to-test

pb/custom_evaluator/custom_evaluator.proto
michelle192837: +1, the other fields are intended to be more generic but since this is using status specifically, I think you can be more specific about what to compare.

michelle192837 (commented):
Thanks Sean! +1 to Sean's comments, otherwise I think this looks good (though note that CustomTargetStatus() isn't currently used anywhere, so that'll need a follow-up in this PR or in a subsequent one before it actually changes any statuses. If that's intended, then all good).

pb/custom_evaluator/custom_evaluator.proto
XuanWang-Amos: Currently only target status is used, looks like we don't have plans to add other fields, I'll change it to `bool`.

XuanWang-Amos:
> Thanks Sean! +1 to Sean's comments, otherwise I think this looks good (though note that CustomTargetStatus() isn't currently used anywhere, so that'll need a follow-up in this PR or in a subsequent one before it actually changes any statuses. If that's intended, then all good).

We're only use it for Sponge test, I'll raise a follow up CL after this merged.

XuanWang-Amos:
@chases2 @michelle192837 PTAL.

pb/custom_evaluator/custom_evaluator.proto
```diff
@@ -52,6 +52,10 @@ message TestResultComparison {
     //
     // NOTE: Only supported for string and numerical values
     string test_result_error_field = 4;
+
+    // This will enable the target status comparation. The value of the status
+    // will be used to evaluate.
+    bool target_status_field = 5;
```
michelle192837, line 58: nit: maybe `target_status` instead since this isn't the name of a field, but the field itself

michelle192837 (approved):
Sorry for the delay! Aside from the one comment, looks good to me.

/hold

Can run `/hold cancel` once the comment is addressed to let this merge. ^^

pb/custom_evaluator/custom_evaluator.proto
XuanWang-Amos: Thanks for the review, will change it in next commit!

XuanWang-Amos:
Looks like `lgtm` label was removed because of new changes.
Sorry for pinging again, @michelle192837 can you help approve again?

michelle192837:
This will need a new release + import, which I'll try and handle by the end of day today.

PR #1161, opened 2023-03-14
WIP: Add a basic ResultStore result source.

michelle192837:
Adds a basic outline of a ResultStore result source, as well as the logic for translating a simple query to a complex query appropriate for the ResultStore API, and validation to ensure a specified simple query is valid. (Specifying this source will not work yet because it's not yet hooked into the Updater.)

Note that this (very loosely) references the code in https://github.com/GoogleCloudPlatform/testgrid/tree/master/resultstore, but the code isn't the same. I'm placing it in an updater subdirectory since this is effectively an addition to the Updater to read from this source.


pkg/updater/resultstore/query.go
```diff
@@ -0,0 +1,334 @@
...
+	}
+	return fmt.Sprintf(token, value), nil
+}
+
+var tokenRegex = regexp.MustCompile(`(?i)(?P<token>(?P<key>\S+):(?P<value>".*?"|'.*?'|\S+))|(?P<logic>or|and|not)|(?P<bare>".*?"|'.*?'|\S+)`)
+
+// ConvertReason includes reasons a conversion might fail.
+type ConvertReason int64
+
+const (
```
sultan-duisenbay, line 118: nit: maybe move all const to the beginning of the file? for better readability

pkg/updater/resultstore/query_test.go
```diff
@@ -0,0 +1,663 @@
...
+		t.Run(tc.query, func(t *testing.T) {
+			got, err := ComplexQuery(tc.query)
+			if err == nil {
+				t.Errorf("complexQuery(%q) expected error but got %q", tc.query, got)
+			}
+		})
+	}
+}
+
+type fakePage struct {
```
sultan-duisenbay, line 402: maybe create a separate fake package or add this to an existing fakeclient package

sultan-duisenbay: or add a TODO for that

sultan-duisenbay (commented):
Amazing job, Michelle! A couple of small comments from me. Mostly about the code structure.

More significant suggestions from my side would be:
- Place a fake invocation client into a separate package and just use it in the tests
- Thanks for creating the tests, but I think we could improve it by testing the higher level functions with different test cases. For instance testing `targetSearch` and `invocationSearch` with different input and kind of simultaneously test the smaller functions at the same time (complexQuery, targetToken, etc)

pkg/updater/resultstore/query_test.go
michelle192837: Hmm, we use both the convention of having it in the same package but separate file, and in a separate package, in this repo. Sticking it in the same package but separate file for now if that sounds good,, since it is specific to the ResultStore testing.

pkg/updater/resultstore/query.go
```diff
@@ -0,0 +1,334 @@
...
+	parts := strings.SplitN(token, ":", 2)
+	key := strings.TrimSpace(parts[0])
+
+	if k := strings.ToLower(key); k == "target" || k == "exact_target" {
+		return true
+	}
+	return false
+}
+
+func complexToken(simpleToken string, target bool) (string, error) {
```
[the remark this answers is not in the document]
michelle192837: Not seeing anything official about bools, but seems like 'is' or 'was' is common and it makes sense. Added!

pkg/updater/resultstore/query.go
michelle192837: Done!

michelle192837:
Manually holding because I want to think over some overall design details before merging.

/hold

michelle192837:
Fixed some boilerplate/format issues.

/lint

PR #1175, opened 2023-03-27
Update npm dependencies.

michelle192837:
Updated the versions for a bunch of dependencies.

Verified `npm run test` and `npm run storybook` still work.

ETA: Most of these are just version updates, but I did replace rollup-plugin-terser with the maintained plugin (@rollup/plugin-terser)


sultan-duisenbay (approved):
LGTM!

PR #1176, opened 2023-04-06
Clicking on any dashboard will route to /dashboards page.

sultan-duisenbay:


web/src/testgrid-router.ts
michelle192837, line 1: Not sure how it'd look, but is it possible to add testing for testgrid-router and/or navigation.ts?

michelle192837 (commented):
Thanks! Have a few comments ^^

sultan-duisenbay:
Thanks for the comments! I addressed those in the upcoming PR

PR #1180, opened 2023-04-07
Add npm presubmit job.

### LINKED ISSUE #1178: Add npm test as a presubmit
Missed an issue with a recent PR that should've been caught if npm test was run as part of the presubmits (see #1175)

/assign

michelle192837:
Kinda jank, but this does successfully run the presubmits.

Forces running on Playwright and uses one of their available images (see https://playwright.dev/docs/ci).

Fixes #1178

/hold

### DIFF
--- a/.prow.yaml
+++ b/.prow.yaml
@@ -12,3 +12,16 @@ presubmits:
       - test
       - --config=ci
       - //...
+- name: test-testgrid-npm
+  decorate: true
+  always_run: true
+  spec:
+    serviceAccountName: presubmits
+    containers:
+    # See https://playwright.dev/docs/ci
+    - image: mcr.microsoft.com/playwright:v1.32.0-focal
+      command:
+      - bash
+      args:
+      - -c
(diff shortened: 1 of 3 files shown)


chases2 (approved):
Looks good! Thanks so much. Unhold when you're ready. :D

bweston92:
Would it be possible to pass the Docker flag for `--ipc=host` or is that a risk?

> Using --ipc=host is also recommended when using Chromium—without it Chromium can run out of memory and crash

https://playwright.dev/docs/ci#docker

michelle192837:
> Would it be possible to pass the Docker flag for `--ipc=host` or is that a risk? Apparently helps with Chromium not running out of memory which make cause flakes.
> 
> > Using --ipc=host is also recommended when using Chromium—without it Chromium can run out of memory and crash
> 
> https://playwright.dev/docs/ci#docker

Good point! I don't think it's a significant risk, but just in case I'm going to leave it off for now, and add it if we see issues with this in presubmit. Thanks!

PR #1185, opened 2023-04-20
Add the tab bar with tab names

sultan-duisenbay:
Add the tab bar (top of the screenshot) - 
Screenshot 2023-04-20 at 5 05 41 PM

### Notable changes
- Added a tab bar with tab names based on mwc components (`mwc-tab` & `mwc-tab-bar`)
- Added the test to make sure tab bar is rendered correctly

### DIFF
--- a/web/test/dashboard-summary.test.ts
+++ b/web/test/dashboard-summary.test.ts
@@ -39,4 +39,19 @@ describe('Testgrid Dashboard Summary page', () => {
 
     expect(element.tabSummariesInfo).to.not.be.empty;
   });
+
+  it('renders the tab bar with tabs', async () => {
+
+    // waiting until list items (dashboards and groups) are fully rendered
+    await waitUntil(
+      () => element.shadowRoot!.querySelector('mwc-tab'),
+      'Dashboard summary did not render tabs within the tab bar',
+      {
+        timeout: 4000,
+      },
+    );
+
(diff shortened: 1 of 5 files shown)


web/test/dashboard-summary.test.ts
```diff
@@ -39,4 +39,19 @@ describe('Testgrid Dashboard Summary page', () => {
...
+    await waitUntil(
+      () => element.shadowRoot!.querySelector('mwc-tab'),
+      'Dashboard summary did not render tabs within the tab bar',
+      {
+        timeout: 4000,
+      },
+    );
+
+    expect(element.tabNames).to.not.be.empty;
+    expect(element.tabNames.length).to.not.equal(1);
```
chases2, line 55: What data are we using to run these unit tests? For a test, we should know how many tab names there are and assert that number.

chases2 (commented):
Looks good overall, but I have one question about how we're unit testing these components.

web/test/dashboard-summary.test.ts
chases2: Please add that as a TODO, either in the code or in a bug. "Creating a mock API" is a substantial, parallel task.

sultan-duisenbay: Got it , thanks!

PR #1186, opened 2023-04-21
Fix to show storybook stories instead of app index.

michelle192837:
Running `npm run storybook` for me produces the TestGrid index page, instead of the Storybook UI with all the stories in the sidebar. I believe it's because we're inheriting the app index from the base config; explicitly ignore that for the storybook server.


sultan-duisenbay (approved):
LGTM!

PR #1189, opened 2023-04-26
Add a tab bar with proper navigation and grid data placeholder.

sultan-duisenbay:
Add a proper navigation to a tab bar and grid data placeholder. Improved the comments and addressed some nits.

### Major changes
- Renamed `dashboard-summary` => `testgrid-dashboard-summary`
- Created a new `testgrid-data-content` component which will render a dashboard summary or grid data based on which tab is clicked on the tab bar.
- Added a `testgrid-grid-display` component responsible for fetching the grid data and rendering it
- Clicking on any tab in the tab bar will also change the URL pathname to /{dashboard} or /{dashboard}/{tab}
- Going to a particular URL will now highlight the correct tab in the tab bar

In the next episode:
- Add the tests and update storybooks for this change


web/src/testgrid-data-content.ts
```diff
@@ -0,0 +1,201 @@
...
+      return
+    }
+    var index = this.tabNames.indexOf(tabName);
+    if (index > -1){
+      this.activeIndex = index;
+    }
+  }
+
+  // fetch the tab data
+  private async fetchTabGridRows() {
```
michelle192837, line 173: qq: imo the grid component should be the one handling the API call, and navigation/routing should just know 1) to get a grid component, and 2) the name of the grid to get (passed to the grid component on init)? I may be wrong on this, but in that case should the 'grid' part of navigation just show the info it'd pass for now? (e.g. the dashboard + tab name).

web/src/testgrid-data-content.ts
michelle192837, line 1: Do we need a <testgrid-data-content>? (I might be overthinking it, but I feel like the rendering for the tab bar is too deep in the internals of the summary and grid rendering atm; is this necessary or can they act more independently?)

michelle192837 (commented):
Thanks! Have a couple structural comments ^^

web/src/testgrid-data-content.ts
sultan-duisenbay: That's a great suggestion and I agree 100%. I will make those changes

web/src/testgrid-data-content.ts
sultan-duisenbay: So the the tab bar is shared between both the dashboard summary and the grid display. In order to provide an interrupted flow of changing the tabs when any tab is clicked, I decided to put it as a common component and then render the internals (dashboard or grid view based on `showTab` variable). We can create a separate `<testgrid-tab-bar>` component if necessary but IMO it will still need to be within a shared container element which encapsulates both dashboard summary and grid data elements. PLS let me know if this answers your question

sultan-duisenbay:
Added the tests just now

michelle192837 (approved):

PR #1190, opened 2023-04-26
Return all cells in API with proper run-length decoding.

michelle192837:
Otherwise, we only return as many cells as there are non-blank cells, which artificially limits the number of cells returned for each row.

Adjusts the index used for messages/cell IDs/icons, since these are only filled in once per non-blank status.  (see https://github.com/GoogleCloudPlatform/testgrid/blob/358f270ad26e8a270cafebfdb8098aa72b54b533/pkg/updater/updater.go#L1200).


sultan-duisenbay (approved):
LGTM! this should help some of the users with explicit null cell results

PR #1191, opened 2023-05-02
Configure API host as env variables.

sultan-duisenbay:
Lets us use different API URLs for dev, test, prod in the future

### DIFF
--- a/web/web-dev-server.config.mjs
+++ b/web/web-dev-server.config.mjs
@@ -1,5 +1,10 @@
 // import { hmrPlugin, presets } from '@open-wc/dev-server-hmr';
 
+import { fromRollup } from '@web/dev-server-rollup';
+import rollupReplace from '@rollup/plugin-replace';
+
+const replace = fromRollup(rollupReplace);
+
 /** Use Hot Module replacement by adding --hmr to the start command */
 const hmr = process.argv.includes('--hmr');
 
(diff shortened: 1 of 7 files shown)


web/web-dev-server.config.mjs
```diff
@@ -18,8 +23,10 @@ export default /** @type {import('@web/dev-server').DevServerConfig} */ ({
   appIndex: 'index.html',
 
   plugins: [
-    /** Use Hot Module Replacement by uncommenting. Requires @open-wc/dev-server-hmr plugin */
-    // hmr && hmrPlugin({ exclude: ['**/*/node_modules/**/*'], presets: [presets.litElement] }),
+    replace({
+      'process.env.API_HOST': '"testgrid-data.k8s.io"',
```
chases2, line 27: One nit: consider adding something in the README about local development that points to this file. That's how I'd run the UI against a local copy of the API, right?

chases2 (approved):
This looks like an excellent development improvement to me. Thanks!
/lgtm
/hold for nits

web/web-dev-server.config.mjs
sultan-duisenbay: In the future, yes. README update sounds great!

sultan-duisenbay: I will add it as a TODO

PR #1192, opened 2023-05-04
Add a basic grid component.

michelle192837:
Adds a basic grid component, consisting of a grid, composed of rows, composed of a row header and cells representing statuses.

(this is experimental and not a final component, but at least some of these components should be reusable even if we switch implementations).

TODO:
- [x] Basic unit tests
- [x] Column headers
- Virtualization (only load visible DOM); follow-up PR


web/src/testgrid-grid-display.ts
```diff
@@ -2,6 +2,7 @@ import { LitElement, html, PropertyValues } from "lit";
 import { map } from "lit/directives/map.js";
 import { customElement, property, state } from "lit/decorators.js";
 import { ListRowsResponse, ListRowsResponse_Row } from './gen/pb/api/v1/data.js';
+import '../src/testgrid-grid-row';
```
sultan-duisenbay, line 5: `./testgrid-grid-row`

web/src/testgrid-grid-row.ts
```diff
@@ -0,0 +1,41 @@
+import { LitElement, html, css } from "lit";
+import { map } from "lit/directives/map.js";
+import { customElement, property } from "lit/decorators.js";
+import { ListRowsResponse_Row } from './gen/pb/api/v1/data.js';
+import { TestStatus } from '../src/gen/pb/test_status/test_status';
+
+interface Cell {
```
sultan-duisenbay, line 7: would it be better to define this interface in the cell component and import it here?

web/src/testgrid-grid-row.ts
```diff
@@ -0,0 +1,41 @@
...
+        ${map(this.rowData.cells,
+        (cell) => html`<testgrid-grid-cell .icon="${cell.icon}" .status="${TestStatus[cell.result]}"></testgrid-grid-cell>`
+      )}
+        `;
+    }
+    return html`
+    <testgrid-grid-row-header .name="${this.name}"></testgrid-grid-row-header>
+    `;
+  }
+}
```
sultan-duisenbay, line 41: I always forget to add a newline too :)

web/src/testgrid-grid-row-header.ts
```diff
@@ -0,0 +1,33 @@
...
+      min-width: 300px;
+    }
+  `;
+
+    @property() name: String;
+
+    render(){
+        return html`${this.name}`;
+    }
+}
```
sultan-duisenbay, line 33: newline here too

web/src/testgrid-grid-cell.ts
```diff
@@ -0,0 +1,68 @@
...
+      font-weight: bold;
+      display: flex;
+      justify-content: center;
+      align-content: center;
+      flex-direction: column;
+      box-sizing: border-box;
+      font-size: 12px;
+    }
+
+    :host([status="NO_RESULT"]) {
```
sultan-duisenbay, line 27: may be assign different classes to statuses... this way we will define common css styles, import them where necessary and use across cell colors and tab summary colors

web/stories/testgrid-grid-display.stories.ts
```diff
@@ -0,0 +1,26 @@
...
+}
+
+const Template: Story<Args> = ({
+  dashboardName = '', tabName = '',
+}: Args) => {
+  return html`<testgrid-grid-display .dashboardName="${dashboardName}" .tabName="${tabName}"></testgrid-grid-display>`;
+};
+
+export const Grid = Template.bind({});
+Grid.args = {dashboardName: 'sig-testing-misc', tabName: 'branchprotector'}
```
sultan-duisenbay, line 26: once we have local API set up, we will have an option of faking those

web/src/testgrid-grid-row.ts
michelle192837: Actually just deleted it, since it's not used currently.

web/src/testgrid-grid-row.ts
```diff
@@ -0,0 +1,41 @@
+import { LitElement, html, css } from "lit";
+import { map } from "lit/directives/map.js";
+import { customElement, property } from "lit/decorators.js";
+import { ListRowsResponse_Row } from './gen/pb/api/v1/data.js';
+import { TestStatus } from '../src/gen/pb/test_status/test_status';
```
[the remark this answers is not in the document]
michelle192837: Done!

web/src/testgrid-grid-display.ts
michelle192837: Done!

web/src/testgrid-grid-row.ts
michelle192837: Ah, thanks! Added an extension in VS Code to just auto-do this from now on 😅

web/src/testgrid-grid-cell.ts
michelle192837: Will consider for refactor, but there's a few things I want to fix more holistically in a follow-up after I check some more resources. Filed #1205 instead if that works!

web/stories/testgrid-grid-display.stories.ts
michelle192837: Thanks for the fake! Working on making this work atm (rest of the PR should be okay for review).

michelle192837: Thanks for the help! Fake works now (I have no idea how I was overcomplicating it 😅 )

web/src/testgrid-grid.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  /**
+   * Lit-element lifecycle method.
+   * Invoked on each update to perform rendering tasks.
+   */
+  render() {
+    return html`
+    <testgrid-grid-header-row .headers="${this.tabGridHeaders}"></testgrid-grid-header-row>
+    ${map(this.tabGridRows,
+      (row: ListRowsResponse_Row) => html`<testgrid-grid-row .name="${row.name}" .rowData="${row}"></testgrid-grid-row>`
```
sultan-duisenbay, line 45: If we are passing the whole row data anyway (which already contains name), .name property can be omitted

web/src/testgrid-grid-row.ts
```diff
@@ -0,0 +1,41 @@
+import { LitElement, html, css } from "lit";
+import { map } from "lit/directives/map.js";
+import { customElement, property } from "lit/decorators.js";
+import { ListRowsResponse_Row } from './gen/pb/api/v1/data.js';
+import { TestStatus } from './gen/pb/test_status/test_status';
+
+interface Cell {
```
sultan-duisenbay, line 7: I believe this was meant to be deleted

web/src/testgrid-grid-row.ts
```diff
@@ -0,0 +1,41 @@
...
+      gap: 0px 2px;
+      margin: 2px;
+    }
+  `;
+
+  @property() name: String;
+  @property() rowData: ListRowsResponse_Row;
+
+  render() {
+    if (this.rowData && this.rowData.cells) {
```
sultan-duisenbay, line 29: here we could make it a little bit more DRY by always defining `testgrid -grid-row-header` first and then add cells if there's data. Also name is an unnecessary property since each row already has `name` field

web/src/testgrid-grid-header-row.ts
```diff
@@ -0,0 +1,34 @@
+import { LitElement, html, css } from "lit";
```
sultan-duisenbay, line 1: IMO `testgrid-grid-header-row` and `testgrid-grid-row-header` are a little bit confusing. perhaps changing the naming would help with that

web/src/testgrid-grid-row-header.ts
```diff
@@ -0,0 +1,32 @@
+import { LitElement, html, css } from "lit";
```
sultan-duisenbay, line 1: IMO `testgrid-grid-header-row` and `testgrid-grid-row-header` are a little bit confusing. perhaps changing the naming would help with that.

how about `testgrid-grid-row-name` or `row-id`?

web/src/testgrid-grid-header-row.ts
```diff
@@ -0,0 +1,34 @@
...
+      margin: 2px;
+    }
+  `;
+
+  @property() headers: ListHeadersResponse;
+
+  render() {
+    if (this.headers && this.headers.headers) {
+      return html`
+        <testgrid-grid-row-header></testgrid-grid-row-header>
```
sultan-duisenbay, line 24: so `testgrid-grid-row-header` is just to move this element to the right by the required padding?
in that case, also we could make it DRY by always setting a `testgrid-grid-row-header` and then checking if headers exist

IMO, would also be great to add a TODO to set the padding / margin in this element's instead of using a placeholder element - to place the element correctly

web/test/testgrid-grid-cell.test.ts
```diff
@@ -0,0 +1,46 @@
+import {
```
sultan-duisenbay, line 1: IMO, for this and column-header, header-row, row-header and row tests cases should be integrated into the `testgrid-grid.test.ts` below to holistically test the grid.

this will lead to less code and cleaner file structure while still reliably testing all of the necessary subcomponents.
Filing an issue is fine :)

sultan-duisenbay (approved):
Thanks for the PR, Michelle! Left a couple of nits but overall approved to unblock this

sultan-duisenbay:
forgot to /hold 
image

but wanted to mention that screenshots of the grid would also be helpful in the PR description :)

michelle192837:
Thanks! Agree on the comments, I'll make a followup PR today 👍

web/src/testgrid-grid-header-row.ts
michelle192837: Atm it's a placeholder, though I think someone did have an FR that was interesting, suggesting the column header name could go to the left of the column header row to describe what it is. (So date, time, build for the defaults, but the column header name/property we're using for the others. Right now it's a tooltip). That said, even if we do something like that, it shouldn't be the same as a row header 😅

PR #1203, opened 2023-05-10
Add fake API for local dev & tests

sultan-duisenbay:
### Major changes

- Fake API spits out the data defined in `fake-api/data.json` with the routes configured in `fake-api/routes.json`
- Created separate web-dev-server configs (1 using fake API, 1 for prod API)
- Changed the web-test-runner config to use fake API (this should prevent flakes in the other PRs)
- Added the npm run start commands for local and k8s to visualize different data
- Updated README


web/test/testgrid-index.test.ts
```diff
@@ -22,10 +22,6 @@ describe('Testgrid Index page', () => {
     element = await fixture(html`<${tag}></${tag}>`);
   });
 
-  it('passes the a11y audit', async () => {
```
michelle192837, line 25: QQ, any reason for removing this?

michelle192837 (approved):
Looks great, thanks for the change Sultan!

Holding for the one comment, otherwise good to go.

/hold

web/test/testgrid-index.test.ts
michelle192837: Ah, gotcha! Commenting + TODO'ing sounds good to me if that's alright, I think we want to pass the accessibility checks but don't have to do it here.

sultan-duisenbay: Adding TODOs now

PR #1213, opened 2023-05-26
Add link to name in dashboard summary page

bryan-lou:
Allow users to use link and navigate to tab from dashboard summary

### DIFF
--- a/web/src/testgrid-dashboard-summary.ts
+++ b/web/src/testgrid-dashboard-summary.ts
@@ -14,6 +14,7 @@ export interface TabSummaryInfo {
   lastUpdateTimestamp: string;
   lastRunTimestamp: string;
   latestGreenBuild: string;
+  dashboardName: string;
 }
 
 // TODO: define in a shared file (dashboard group also uses this)
@@ -39,6 +40,7 @@ function convertResponse(ts: TabSummary) {
     ).toISOString(),
     lastRunTimestamp: Timestamp.toDate(ts.lastRunTimestamp!).toISOString(),
     latestGreenBuild: ts.latestPassingBuild,
+    dashboardName: ts.dashboardName,
   };
   return tsi;
 }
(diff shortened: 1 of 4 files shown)


web/src/testgrid-dashboard-summary.ts
```diff
@@ -88,6 +90,7 @@ export class TestgridDashboardSummary extends LitElement {
       var tabSummaries: Array<TabSummaryInfo> = [];
       data.tabSummaries.forEach(ts => {
         const si = convertResponse(ts);
+        si.dashboardName = this.dashboardName
```
sultan-duisenbay, line 93: I believe this can be done within `convertResponse` function

web/src/utils/navigation.ts
```diff
@@ -23,4 +23,5 @@ export function navigateTab(dashboard: string, tab: string){
     url.pathname = `${dashboard}/${tab}`
   }
   history.pushState(null, '', url);
+  window.dispatchEvent(new CustomEvent('location-changed'));
```
sultan-duisenbay, line 26: could you create another function or change this one? e.g. add a boolean to control the event dispatch behaviour?

tab bar relies on `navigateTab` to not reload, but for your case, we should.

sultan-duisenbay (commented):
Thanks Bryan! Couple of nits from me

web/src/testgrid-dashboard-summary.ts
bryan-lou: Done.

web/src/utils/navigation.ts
bryan-lou: Done.

web/src/testgrid-dashboard-summary.ts
```diff
@@ -28,7 +29,7 @@ const tabStatusIcon = new Map<string, string>([
 ]);
 
 // TODO: generate the correct time representation
-function convertResponse(ts: TabSummary) {
+function convertResponse(ts: TabSummary, dbName: string) {
```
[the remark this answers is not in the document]
sultan-duisenbay: I believe this was not working correctly because our fake-server was not providing correct dashboard name - the fix is available in https://github.com/GoogleCloudPlatform/testgrid/pull/1216. This is fine for now

sultan-duisenbay (approved):
LGTM! Thanks Bryan.

/hold

PR #1217, opened 2023-05-31
Remove empty results from truncation column.

michelle192837:
The unit tests can imply otherwise, but when we inflate a column for a grid we actually have a cell for each row, which may have a real status _or_ may be empty. Only replace non-empty cells with the truncation warning, so empty rows can be dropped after truncation.


sultan-duisenbay (approved):
LGMT /hold

michelle192837:
Thanks!

/hold cancel

PR #1221, opened 2023-06-09
Create failing tests dropdown component in dashboard tab summary

bryan-lou:


web/src/testgrid-tab-table.ts
```diff
@@ -0,0 +1,61 @@
+import { LitElement, html, css } from 'lit';
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import { customElement, property } from 'lit/decorators.js';
+import {TabSummaryInfo} from './testgrid-dashboard-summary';
+
+@customElement('testgrid-tab-table')
+export class TestgridTabTable extends LitElement {
+  @property() clicked = '';
+  @property() visible = false;
```
sultan-duisenbay, line 9: this should probably be a state and not a property. please refer to https://lit.dev/docs/components/properties/#internal-reactive-state

web/src/testgrid-tab-table.ts
```diff
@@ -0,0 +1,61 @@
...
+  info?: TabSummaryInfo;
```
sultan-duisenbay, line 10: For this particular element we don't need the full tabSummaryInfo, right? We can just provide failuresSummary

web/src/testgrid-tab-table.ts
```diff
@@ -0,0 +1,61 @@
...
+              <th>Last Passed</th>
+            </tr>
+          </table>`
+          : ''}
+      </div>
+    `
+  }
+  private dropdownTable(){
+    this.visible = !this.visible;
+    this.dispatchEvent(new CustomEvent('visibleChange', { detail: this.visible }));
```
sultan-duisenbay, line 33: what is the purpose of this event?

sultan-duisenbay (changes requested):
Thanks for the PR, Bryan! Left comments here and there. The major one is to finalize a separate element for the table data and embed it into `tab-summary` element.

In addition to this, would be great to add it to tab summary stories - make some of them have failuresSummary. Also add the test cases for tab summary

web/src/tab-summary.ts
```diff
@@ -1,11 +1,13 @@
 import { LitElement, html, css } from 'lit';
+import { map } from 'lit/directives/map.js';
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 import { customElement, property } from 'lit/decorators.js';
 import { TabSummaryInfo } from './testgrid-dashboard-summary';
 @customElement('tab-summary')
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 export class TabSummary extends LitElement {
   @property({ type: Object })
+  @property() visible = false;
```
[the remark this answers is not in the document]
bryan-lou: done

web/src/testgrid-dashboard-summary.ts
```diff
@@ -42,6 +58,27 @@ function convertResponse(ts: TabSummary) {
     latestGreenBuild: ts.latestPassingBuild,
     dashboardName: ts.dashboardName,
   };
+  if (ts.failuresSummary !== undefined) {
+    tsi.failuresSummary = {} as FailuresSummaryInfo
+  }
+  if (ts.failuresSummary?.failureStats !== undefined) {
```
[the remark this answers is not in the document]
bryan-lou: done

web/src/testgrid-tab-table.ts
```diff
@@ -0,0 +1,61 @@
+import { LitElement, html, css } from 'lit';
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import { customElement, property } from 'lit/decorators.js';
+import {TabSummaryInfo} from './testgrid-dashboard-summary';
+
+@customElement('testgrid-tab-table')
+export class TestgridTabTable extends LitElement {
+  @property() clicked = '';
```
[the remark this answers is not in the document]
bryan-lou: removed

web/src/testgrid-tab-table.ts
bryan-lou: Yup, that's right, but actually right now I imported the `TabSummaryInfo`, but there's no data in this object when it gets called. So I'm still looking into this and trying to figure out why it's working in tab-summary but not this

web/src/testgrid-tab-table.ts
bryan-lou: why shouldn't we use a property for this?

web/src/testgrid-tab-table.ts
```diff
@@ -0,0 +1,61 @@
+import { LitElement, html, css } from 'lit';
```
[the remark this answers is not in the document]
sultan-duisenbay: testgrid-tab-table

web/src/testgrid-tab-table.ts
sultan-duisenbay: property is an externally set attribute while state is internal. looks like visible is toggled within the table element

web/src/tab-summary.ts
```diff
@@ -39,11 +41,37 @@ export class TabSummary extends LitElement {
           </div>
         </div>
       </div>
+      ${this.info?.failuresSummary !== undefined ? html `<div class="dropdown-container">
```
sultan-duisenbay, line 44: would be great to ember the table element you defined below here

sultan-duisenbay (changes requested):

web/src/testgrid-tab-table.ts
bryan-lou: fixed

web/src/testgrid-tab-table.ts
bryan-lou: Done

web/src/tab-summary.ts
bryan-lou: done

web/src/testgrid-failures-summary.ts
```diff
@@ -0,0 +1,74 @@
+import { LitElement, html, css } from 'lit';
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import { customElement, property, state } from 'lit/decorators.js';
+import { TabSummaryInfo } from './testgrid-dashboard-summary';
+import { map } from 'lit/directives/map.js';
+
+@customElement('testgrid-failures-summary')
+export class TestgridTabTable extends LitElement {
```
sultan-duisenbay, line 8: It's better to keep the class name the same the custom element name. i.e
```
export class TestgridFailuresSummary extends LitElement
```

sultan-duisenbay (changes requested):
One nit with changing class name

web/src/testgrid-failures-summary.ts
bryan-lou: Done!

sultan-duisenbay (approved):
/hold

Please squash the commits (have 1 instead of 7)

PR #1222, opened 2023-06-13
Add ResultStore search to Updater.

michelle192837:
Initial commit that adds some logic and basic unit tests for these functions.

Will follow up with further changes once I verify this works against a real API.


michelle192837:
/assign @chases2 @sultan-duisenbay 

I know this is a long change (apologies!), let me know if I can break it up for easier reviewing. ^^

sultan-duisenbay (approved):
LGTM with couple of nits and questions!

/hold

PR #1228, opened 2023-06-23
WIP: Improve frontend code structure

sultan-duisenbay:
We are creating many components, so it's time to structurize those elements.

Moved al routers to navigation dir, all grid components to grid dir, all dashboard components to dashboard dir and group related components to group dir.


michelle192837:
This is great, thanks! I think it should be separated from the other commit though, or indicate this is blocked by another PR.

PR #1232, opened 2023-06-27
Alert stale tab when no rows exist.

michelle192837:
It's possible to have a grid with a column (see https://github.com/GoogleCloudPlatform/testgrid/blob/6ad903765c1a60491ea661d268780aa0e0de71ea/pkg/tabulator/tabstate.go#L445, which keeps a column if the grid would otherwise be empty) but no rows. This is fine, but the stale alert doesn't account for this at the moment. Update it to alert when no rows exist, in addition to the other reasons a tab might be stale.


sultan-duisenbay (approved):
LGTM!
/hold

michelle192837:
Thanks!
/hold cancel

PR #1236, opened 2023-07-14
Add new pod monitors for metrics on GMP.

michelle192837:
Following https://cloud.google.com/stackdriver/docs/solutions/gke/gmp-migration to migrate us to managed prometheus so our metrics work again ^^


michelle192837:
/hold

Actually, since it's just adding new resources, no reason I can't add canary and prod in the same build.

sultan-duisenbay (approved):

PR #1242, opened 2023-07-25
Extract grouping ID based on configurations

bryan-lou:
Extract the group ID based on primary grouping option or fallback to a default grouping by invocation

### DIFF
--- a/pkg/updater/resultstore/resultstore.go
+++ b/pkg/updater/resultstore/resultstore.go
@@ -70,6 +70,20 @@ type processedResult struct {
 	TargetResults   map[string][]*singleActionResult
 }
 
+// extractGroupID extracts grouping ID for a results based on the testgroup grouping configuration
+// Returns an empty string for no config or incorrect config
+func extractGroupID(tg *configpb.TestGroup, result *processedResult) string {
+	switch {
+	// P - build info
+	case result == nil:
+		return ""
+	case tg.GetPrimaryGrouping() == configpb.TestGroup_PRIMARY_GROUPING_BUILD:
+		return identifyBuild(tg, result)
+	default:
+		return result.InvocationProto.GetId().GetInvocationId()
+	}
+}
+
 // ResultStoreColumnReader fetches results since last update from ResultStore and translates them into columns.
 func ResultStoreColumnReader(client *DownloadClient, reprocess time.Duration) updater.ColumnReader {
 	return func(ctx context.Context, log logrus.FieldLogger, tg *configpb.TestGroup, oldCols []updater.InflatedColumn, defaultStop time.Time, receivers chan<- updater.InflatedColumn) error {
@@ -97,7 +111,6 @@ func ResultStoreColumnReader(client *DownloadClient, reprocess time.Duration) up
 		})
 
 		for _, pr := range processedResults {
-			// TODO: Make group ID something other than start time.
 			inflatedCol := processGroup(tg, pr)
 			receivers <- *inflatedCol
 		}
--- a/pkg/updater/resultstore/resultstore_test.go
+++ b/pkg/updater/resultstore/resultstore_test.go
@@ -93,6 +93,83 @@ func timeMustText(t time.Time) string {
 	return string(s)
 }
 
+func TestExtractGroupID(t *testing.T) {
+	cases := []struct {
+		name string
+		tg   *configpb.TestGroup
+		pr   *processedResult
+		want string
+	}{
+		{
+			name: "nil",
+		},
+		{
+			name: "primary grouping BUILD by override config value",
+			tg: &configpb.TestGroup{
+				DaysOfResults:                   7,
+				BuildOverrideConfigurationValue: "test-key-1",
+				PrimaryGrouping:                 configpb.TestGroup_PRIMARY_GROUPING_BUILD,
+			},
+			pr: &processedResult{
+				InvocationProto: &resultstore.Invocation{
+					Id: &resultstore.Invocation_Id{
+						InvocationId: "id-1",
+					},
+					Properties: []*resultstore.Property{
+						{
+							Key:   "test-key-1",
+							Value: "test-val-1",
+						},
+					},
+					Name: invocationName("id-1"),
+					Timing: &resultstore.Timing{
+						StartTime: &timestamppb.Timestamp{
+							Seconds: 1234,
+						},
+					},
+				},
+			},
+			want: "test-val-1",
+		},
+		{
+			name: "fallback grouping BUILD resort to default",
+			tg: &configpb.TestGroup{
+				DaysOfResults:                   7,
+				BuildOverrideConfigurationValue: "test-key-1",
+				FallbackGrouping:                configpb.TestGroup_FALLBACK_GROUPING_BUILD,
+			},
+			pr: &processedResult{
+				InvocationProto: &resultstore.Invocation{
+					Id: &resultstore.Invocation_Id{
+						InvocationId: "id-1",
+					},
+					Properties: []*resultstore.Property{
+						{
+							Key:   "test-key-1",
+							Value: "test-val-1",
+						},
+					},
+					Name: invocationName("id-1"),
+					Timing: &resultstore.Timing{
+						StartTime: &timestamppb.Timestamp{
+							Seconds: 1234,
+						},
+					},
+				},
+			},
+			want: "id-1",
+		},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			got := extractGroupID(tc.tg, tc.pr)
+			if diff := cmp.Diff(tc.want, got); diff != "" {
+				t.Errorf("extractGroupID() differed (-want, +got): %s", diff)
+			}
+		})
+	}
+}
+
 func TestResultStoreColumnReader(t *testing.T) {
 	// We already have functions testing 'stop' logic.
 	// Scope this test to whether the column reader fetches and returns ascending results.


pkg/updater/resultstore/resultstore.go
```diff
@@ -70,6 +73,117 @@ type processedResult struct {
 	TargetResults   map[string][]*singleActionResult
 }
 
+type timedError struct {
```
sultan-duisenbay, line 76: imo - we don't need this now

pkg/updater/resultstore/resultstore.go
```diff
@@ -70,6 +73,117 @@ type processedResult struct {
...
+	millis int64
+	err    error
+}
+type invocationGroup struct {
+	group       string
+	invocations []processedResult
+	errs        map[string]timedError
+}
+
+const (
```
sultan-duisenbay, line 86: dont need those too

pkg/updater/resultstore/resultstore.go
```diff
@@ -70,6 +73,117 @@ type processedResult struct {
...
+	// PartOf is the configuration value key which links to the group invocation ID.
+	PartOf = "EXECUTED_AS_PART_OF_INVOCATION"
+	// Originally indicates the associated non-overview indication.
+	Originally = "Original_test_invocation"
+)
+
+// buildIdentifier may assume its arguments are non-nil.
+type buildIdentifier func(configpb.TestGroup, processedResult) string
```
sultan-duisenbay, line 94: we already have a function to identifybuilds and can call it directly with required parameters

pkg/updater/resultstore/resultstore.go
```diff
@@ -70,6 +73,117 @@ type processedResult struct {
...
+
+var (
```
sultan-duisenbay, line 96: var () is used with many variables at the same time. also `zero` is only used in a single place

pkg/updater/resultstore/resultstore.go
```diff
@@ -70,6 +73,117 @@ type processedResult struct {
...
+	zero = time.Unix(0, 0)
+)
+
+func timeRunDate(milliseconds int64) time.Time {
+	return zero.Add(time.Duration(milliseconds) * time.Millisecond)
+}
+
+func invocationGroupID(inv processedResult, primary configpb.TestGroup_PrimaryGrouping, fallback configpb.TestGroup_FallbackGrouping, fallbackKey string, identifyBuild string) string {
```
sultan-duisenbay, line 104: to identify groupId we just need test group configuration and processed result. better to use test group directly instead of having all of those individual fields to extract group ID

sultan-duisenbay: we can also identify build using an already existing function which leverages a processed result and test group configuration

pkg/updater/resultstore/resultstore.go
```diff
@@ -70,6 +73,117 @@ type processedResult struct {
...
+	if primary == configpb.TestGroup_PRIMARY_GROUPING_BUILD {
+		return identifyBuild
```
sultan-duisenbay, line 106: leverage the idenitfybuild function here

pkg/updater/resultstore/resultstore.go
```diff
@@ -70,6 +73,117 @@ type processedResult struct {
...
+	}
+	// PRIMARY_GROUPING_NONE, consider fallback
+	id := inv.InvocationProto.GetId().String()
```
sultan-duisenbay, line 109: let's remove this - imo we don't need this here... we can add this later if need be. let's focus on primary and fallback groupings. only

sultan-duisenbay, line 109: will this give you invocationID

pkg/updater/resultstore/resultstore.go
```diff
@@ -70,6 +73,117 @@ type processedResult struct {
...
+		}
+		parts := strings.Split(cfg.GetValue(), "/")
+		if groupID := parts[len(parts)-1]; groupID != id {
+			return groupID
+		}
+		break
+	}
+	switch fallback {
+	case configpb.TestGroup_FALLBACK_GROUPING_DATE:
+		return timeRunDate(inv.InvocationProto.GetTiming().StartTime.GetSeconds()).UTC().Format("gid-2006-01-02")
```
sultan-duisenbay, line 123: GetStartTime()

pkg/updater/resultstore/resultstore.go
```diff
@@ -70,6 +73,117 @@ type processedResult struct {
...
+	case configpb.TestGroup_FALLBACK_GROUPING_LABELS:
+		h := fnv.New32a()
+		for _, l := range inv.InvocationProto.InvocationAttributes.GetLabels() {
+			h.Write([]byte(l))
+		}
+		return "gid-" + strconv.Itoa(int(h.Sum32()))
+	case configpb.TestGroup_FALLBACK_GROUPING_ID:
+		return id
+	case configpb.TestGroup_FALLBACK_GROUPING_BUILD:
+		return "gid-" // FIX THIS+ inv.cl // true even if overriding changelist
```
sultan-duisenbay, line 133: you can use identify build function here

pkg/updater/resultstore/resultstore.go
```diff
@@ -70,6 +73,117 @@ type processedResult struct {
...
+			}
+		}
+		return ""
+	default: // FALLBACK_GROUPING_NONE
+		return ""
+	}
+}
+
+// Returns {groupID: [invocation, invocaton]}, aka the details of all invocations belonging in a group
+func groupInvocations(ctx context.Context, log logrus.FieldLogger, tg *configpb.TestGroup, processedResults []*processedResult, identifyBuild buildIdentifier, invocationGroups chan<- *invocationGroup, receivers chan<- updater.InflatedColumn) error {
```
sultan-duisenbay, line 147: this function is not used anywhere

pkg/updater/resultstore/resultstore.go
```diff
@@ -96,8 +210,23 @@ func ResultStoreColumnReader(client *DownloadClient, reprocess time.Duration) up
 			return processedResults[i].InvocationProto.GetTiming().GetStartTime().GetSeconds() > processedResults[j].InvocationProto.GetTiming().GetStartTime().GetSeconds()
 		})
 
+		groups := make(map[string]*invocationGroup, 1)
+
+		updateGroup := func(inv processedResult) *invocationGroup {
```
sultan-duisenbay, line 215: i would prefer to use regular functions over closures with comments - look at `processRawResults` and other functions

pkg/updater/resultstore/resultstore.go
```diff
@@ -96,8 +210,23 @@ func ResultStoreColumnReader(client *DownloadClient, reprocess time.Duration) up
...
+				group = &invocationGroup{
+					group: groupID,
+				}
+				groups[groupID] = group
+			}
+			return group
+		}
 		for _, pr := range processedResults {
-			// TODO: Make group ID something other than start time.
+			group := updateGroup(*pr)
```
sultan-duisenbay, line 228: we are not processing groups

pkg/updater/resultstore/resultstore.go
bryan-lou: > to identify groupId we just need test group configuration and processed result. better to use test group directly instead of having all of those individual fields to extract group ID

got it

pkg/updater/resultstore/resultstore.go
bryan-lou: ok, Sounds good

pkg/updater/resultstore/resultstore.go
bryan-lou: removed it for now

pkg/updater/resultstore/resultstore.go
bryan-lou: okay

pkg/updater/resultstore/resultstore.go
```diff
@@ -70,6 +72,48 @@ type processedResult struct {
 	TargetResults   map[string][]*singleActionResult
 }
 
+type invocationGroup struct {
```
sultan-duisenbay, line 75: you can remove this for now

pkg/updater/resultstore/resultstore_test.go
```diff
@@ -93,6 +95,176 @@ func timeMustText(t time.Time) string {
...
+						StartTime: &timestamppb.Timestamp{
+							Seconds: 1234,
+						},
+					},
+				},
+			},
+			want: "gid-1970-01-01",
+		},
+		{
+			name: "fallback grouping LABELS",
```
sultan-duisenbay, line 158: I would add some labels to this test case, and set the `want` to some static string like "gid-11233141244"

sultan-duisenbay (approved):
Largely LGTM with a couple of nits, please address those and it's good to be merged!

/hold

bryan-lou:
> Largely LGTM with a couple of nits, please address those and it's good to be merged!
> 
> /hold

Addressed comments

pkg/updater/resultstore/resultstore.go
michelle192837: This depends imo. If we're using a lot of fields from the test group then I think it makes sense, but if we need only one or two, then just passing the fields often makes testing and understanding what the function depends on a lot easier.

pkg/updater/resultstore/resultstore.go
```diff
@@ -70,6 +72,43 @@ type processedResult struct {
...
 
+// extractGroupID extracts grouping ID for a results based on the testgroup grouping configuration
+// Returns an empty string for no config or incorrect config
+func extractGroupID(tg *configpb.TestGroup, result *processedResult) string {
+	switch {
+	// P - build info
+	case tg.GetPrimaryGrouping() == configpb.TestGroup_PRIMARY_GROUPING_BUILD:
+		return identifyBuild(tg, result)
+	// F - date (YYYY-mm-dd)
+	case tg.GetFallbackGrouping() == configpb.TestGroup_FALLBACK_GROUPING_DATE:
```
michelle192837, line 83: nit: imo we shouldn't include fallback_grouping at all in processing, we should just set a reasonable default for ResultStore. Users shouldn't need to set this usually, we should be able to figure it out by default based on the result source.

sultan-duisenbay: do you suggest to use primary grouping and then if nothing set, resort to grouping by invocation ID?

michelle192837: Yup. I don't think fallback_grouping will help for this result source; if we need it later we can introduce it again.

bryan-lou: I just removed fallback_grouping. Can you take a look again? @sultan-duisenbay @michelle192837

pkg/updater/resultstore/resultstore.go
```diff
@@ -79,31 +77,6 @@ func extractGroupID(tg *configpb.TestGroup, result *processedResult) string {
...
-	// F - configuration value
-	case tg.GetFallbackGrouping() == configpb.TestGroup_FALLBACK_GROUPING_CONFIGURATION_VALUE:
-		for _, property := range result.InvocationProto.GetProperties() {
-			if property.GetKey() == tg.GetFallbackGroupingConfigurationValue() {
-				return property.GetValue()
-			}
-		}
-		return ""
 	default:
 		return ""
```
sultan-duisenbay, line 81: for default behaviour here we need to return invocationID
`result.InvocationProto.GetId().GetInvocationId()`

sultan-duisenbay (changes requested):
small change required

pkg/updater/resultstore/resultstore.go
bryan-lou: done.

pkg/updater/resultstore/resultstore_test.go
```diff
@@ -93,6 +95,61 @@ func timeMustText(t time.Time) string {
...
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			got := extractGroupID(tc.tg, tc.pr)
+			if tc.name == "fallback grouping LABELS" {
+				h := fnv.New32a()
+				h.Write([]byte("test-label-1"))
+				tc.want = tc.want + strconv.Itoa(int(h.Sum32()))
+			}
+			strconv.Itoa(int((fnv.New32a().Sum32())))
```
sultan-duisenbay, line 145: we are not using fallback grouping by labels - so no need for those

pkg/updater/resultstore/resultstore_test.go
```diff
@@ -93,6 +95,61 @@ func timeMustText(t time.Time) string {
...
+					Name: invocationName("id-1"),
+					Timing: &resultstore.Timing{
+						StartTime: &timestamppb.Timestamp{
+							Seconds: 1234,
+						},
+					},
+				},
+			},
+			want: "test-val-1",
+		},
```
sultan-duisenbay, line 135: please add a test case for any fallback grouping which will default to invocation ID

pkg/updater/resultstore/resultstore_test.go
bryan-lou: good call - did not catch that

sultan-duisenbay (changes requested):

sultan-duisenbay (approved):

PR #1246, opened 2023-08-02
Group invocations according to grouping strategy

sultan-duisenbay:
Added the functionality to group the results based on the group ID in #1242.
Added unit tests for the functionality above.
Renamed some variables for better clarity

### DIFF
--- a/pkg/updater/resultstore/resultstore.go
+++ b/pkg/updater/resultstore/resultstore.go
@@ -65,22 +65,24 @@ type multiActionResult struct {
 	ActionProtos          []*resultstorepb.Action
 }
 
-type processedResult struct {
+// invocation is an internal invocation representation which contains
+// actual invocation data and results for each target
+type invocation struct {
 	InvocationProto *resultstorepb.Invocation
 	TargetResults   map[string][]*singleActionResult
 }
 
 // extractGroupID extracts grouping ID for a results based on the testgroup grouping configuration
 // Returns an empty string for no config or incorrect config
-func extractGroupID(tg *configpb.TestGroup, result *processedResult) string {
+func extractGroupID(tg *configpb.TestGroup, inv *invocation) string {
 	switch {
 	// P - build info
-	case result == nil:
+	case inv == nil:
 		return ""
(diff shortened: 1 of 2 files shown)


pkg/updater/resultstore/resultstore.go
```diff
@@ -103,38 +105,76 @@ func ResultStoreColumnReader(client *DownloadClient, reprocess time.Duration) up
...
+			groupedInvocations[groupIdentifier] = group
+		}
+		group.Invocations = append(group.Invocations, invocation)
+	}
+
+	for _, group := range groupedInvocations {
+		sortedGroups = append(sortedGroups, group)
+	}
+
+	// reverse sort groups by invocation time
```
michelle192837, line 153: nit: I think sorting in the function is a bit opaque (especially if the function doc doesn't clarify it's sorting). imo sorting should happen more local to where it's needed, or this function should clarify it sorts the invocations.

pkg/updater/resultstore/resultstore_test.go
```diff
@@ -1359,3 +1359,504 @@ func TestIncludeStatus(t *testing.T) {
 		})
 	}
 }
+
+func TestGroupInvocations(t *testing.T) {
+	cases := []struct {
+		name        string
+		tg          *configpb.TestGroup
+		invocations []*invocation
+		want        []*invocationGroup
```
michelle192837, line 1368: nit: Does it make sense to make the 'want' argument more compact? I think all this cares about is that each group ID has the right invoc IDs, and I think it'll be more readable.

michelle192837 (approved):
Mostly LGTM aside from some comments. Thanks!

/hold

Unhold at will

pkg/updater/resultstore/resultstore.go
sultan-duisenbay: Added a comment to the function

pkg/updater/resultstore/resultstore_test.go
sultan-duisenbay: Good suggestion - I made the tests more compact by removing some parts which didn't matter

PR #1247, opened 2023-08-02
Process groups instead of individual invocations

sultan-duisenbay:
Process invocation groups - aka convert each group into its own InflatedColumn.
Invocation groups were created in #1246 
Added unit tests.

### DIFF
--- a/pkg/updater/resultstore/resultstore.go
+++ b/pkg/updater/resultstore/resultstore.go
@@ -112,8 +112,9 @@ func ResultStoreColumnReader(client *DownloadClient, reprocess time.Duration) up
 			return invocations[i].InvocationProto.GetTiming().GetStartTime().GetSeconds() > invocations[j].InvocationProto.GetTiming().GetStartTime().GetSeconds()
 		})
 
-		for _, inv := range invocations {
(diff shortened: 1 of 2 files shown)


pkg/updater/resultstore/resultstore.go
```diff
@@ -279,57 +281,71 @@ func includeStatus(tg *configpb.TestGroup, sar *singleActionResult) bool {
...
-				status = statuspb.TestStatus_UNKNOWN
+	groupedCells := make(map[string][]updater.Cell)
+
+	var hint int64
+
+	// extract info from underlying invocations and target results
+	for _, invocation := range group.Invocations {
+
+		if build := identifyBuild(tg, invocation); build != "" {
+			col.Column.Build = build
```
michelle192837, line 304: What happens if invocations produce different build IDs, but have a consistent group ID?

sultan-duisenbay: IIUC, build maintains the latest build value of the group (at least in the sponge code)

sultan-duisenbay: Discussed in a separate channel :)

michelle192837 (approved):

PR #1248, opened 2023-08-04
Enable custom target result status

sultan-duisenbay:
We are just reusing the functionality within updater package, which is already unit tested

### DIFF
--- a/pkg/updater/resultstore/resultstore.go
+++ b/pkg/updater/resultstore/resultstore.go
@@ -28,11 +28,15 @@ import (
 	"github.com/sirupsen/logrus"
 
 	configpb "github.com/GoogleCloudPlatform/testgrid/pb/config"
+	cepb "github.com/GoogleCloudPlatform/testgrid/pb/custom_evaluator"
 	statepb "github.com/GoogleCloudPlatform/testgrid/pb/state"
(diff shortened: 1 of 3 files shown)


michelle192837 (commented):
This looks good, but I think this should include where we're adding customTargetStatus().

sultan-duisenbay:
> This looks good, but I think this should include where we're adding customTargetStatus().

Will do

pkg/updater/resultstore/resultstore.go
```diff
@@ -59,6 +63,12 @@ type singleActionResult struct {
 	ActionProto           *resultstorepb.Action
 }
 
+// make singleActionResult satisfy TargetResult interface
+func (sar *singleActionResult) TargetStatus() statuspb.TestStatus {
+	status := convertStatus[sar.TargetProto.GetStatusAttributes().GetStatus()]
```
chases2, line 68: nit: one-line return would be just as clear

chases2 (approved):

pkg/updater/resultstore/resultstore.go
chases2: Oh no I forgot to hold :(

PR #1251, opened 2023-08-29
add functionality to extract custom headers

sultan-duisenbay:
Headers can be extracted from both invocations as well as single action target results

### DIFF
--- a/pkg/updater/resultstore/resultstore.go
+++ b/pkg/updater/resultstore/resultstore.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"fmt"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/GoogleCloudPlatform/testgrid/pkg/updater"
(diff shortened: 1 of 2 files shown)


pkg/updater/resultstore/resultstore.go
```diff
@@ -294,6 +337,30 @@ func includeStatus(tg *configpb.TestGroup, sar *singleActionResult) bool {
...
+	var properties []*resultstorepb.Property
+	for _, p := range t.suiteProto.GetProperties() {
+		properties = append(properties, p)
+	}
+	for _, p := range t.caseProto.GetProperties() {
+		properties = append(properties, p)
+	}
+
+	for _, t := range t.suiteProto.GetTests() {
+		newTestResult := &testResult{t.GetTestSuite(), t.GetTestCase()}
```
chases2, line 358: nit: I don't think the `&` does anything here?

chases2 (approved):
Looks good from a readability stance; will let someone else comment if desired.
/hold

pkg/updater/resultstore/resultstore.go
sultan-duisenbay: i think you're right - you can provide both a pointer and variable to a pointer receiver method

PR #1252, opened 2023-09-06
Add properties, and a few stats to testsuite, and status to testcase.

minzhuogoogle:
Add properties for testsuite, along with a few other fields.

Properties, disabled, errors, and status are set for testsuite. And properties have critical information related with testsuite as name in the majority of junit files. For stats disabled, skipped and etc are numbers needed to be extracted too.

### DIFF
--- a/metadata/junit/junit.go
+++ b/metadata/junit/junit.go
@@ -69,7 +69,7 @@ type Suite struct {
 	TimeStamp  string      `xml:"timestamp,omitempty"`
 	Failures   int         `xml:"failures,attr"`
 	Tests      int         `xml:"tests,attr"`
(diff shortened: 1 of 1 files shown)


michelle192837 (commented):
Thanks! New fields look good, I think another couple test cases (or extensions of the existing ones in TestParse() in junit_test.go) with the new fields will be good. 👍

michelle192837:
> Thanks! New fields look good, I think another couple test cases (or extensions of the existing ones in TestParse() in junit_test.go) with the new fields will be good. 👍

Ah, and in addition to unit tests, if you can squash the commits down into 1 commit that would be great. ^^

PR #1254, opened 2023-09-19
Insert icon and message to inflated column cell

bryan-lou:


pkg/updater/resultstore/resultstore.go
```diff
@@ -174,6 +175,118 @@ func ResultStoreColumnReader(client *DownloadClient, reprocess time.Duration) up
...
+func parseTaxonomizer(msg string) (string, string) {
+	parts := strings.SplitN(msg, "~", 2) // class_type~message
+	if len(parts) != 2 {
+		return "", ""
+	}
+	return parts[0], parts[1]
+}
+
+// cellMessageIcon attempts to find an interesting message and icon by looking at the associated properties and tags.
+func cellMessageIcon(annotations []*configpb.TestGroup_TestAnnotation, properties map[string][]string, tags []string) (string, string) {
```
michelle192837, line 228: Does anything use this function at the moment?

pkg/updater/resultstore/resultstore.go
```diff
@@ -174,6 +175,118 @@ func ResultStoreColumnReader(client *DownloadClient, reprocess time.Duration) up
...
 
+const (
+	passWithErrors   = "pass_with_errors"
+	passWithSkips    = "pass_with_skips"
+	skipReason       = "skip_reason"
+	categorizedAbort = "categorized_abort"
+	categorizedFail  = "categorized_fail"
+	linkedBug        = "linked_bug"
+	blockingBug      = "blocking_bug"
+	taxonomizerClass = "taxonomizer.classification"
```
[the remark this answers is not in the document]
michelle192837: Sorry for the delay. I think this comment is still true; I don't think we want to process any of the well-known properties by default. Instead, we should point people to the more generic rules for status override.

pkg/updater/resultstore/resultstore_test.go
```diff
@@ -488,6 +488,119 @@ func TestResultStoreColumnReader(t *testing.T) {
...
+						PropertyName: "actually-a-tag",
+					},
+				},
+			},
+			tags:    []string{"actually-a-tag"},
+			message: "actually-a-tag",
+			icon:    "icon",
+		},
+		{
+			name: "find from pass_with_errors",
```
michelle192837, line 534: These and the other test cases that reference well-known properties can also be removed.

pkg/updater/resultstore/resultstore_test.go
```diff
@@ -488,6 +488,119 @@ func TestResultStoreColumnReader(t *testing.T) {
...
+		properties  map[string][]string
+		tags        []string
+		message     string
+		icon        string
+	}{
+		{
+			name: "basically works",
+		},
+		{
+			name: "find annotation from property",
```
michelle192837, line 504: I think we want one more test case covering the behavior when there's multiple test annotations specified.

pkg/updater/resultstore/resultstore.go
```diff
@@ -178,6 +179,72 @@ func ResultStoreColumnReader(client *DownloadClient, reprocess time.Duration) up
...
+	checkOrder := make([]string, 0, len(annotations))
+
+	for _, a := range annotations {
+		k := a.GetPropertyName()
+		check[k] = a.ShortText
+		tagMap[k] = a.ShortText
+		checkOrder = append(checkOrder, k)
+	}
+
+	for _, key := range checkOrder {
```
michelle192837, line 208: checkOrder is now just iterating over the annotations in order, so I think the logic above can be simplified.

pkg/updater/resultstore/resultstore.go
```diff
@@ -458,6 +525,12 @@ func processGroup(tg *configpb.TestGroup, group *invocationGroup) *updater.Infla
 						headers[i] = append(headers[i], targetHeaders...)
 					}
 				}
+
+				properties := map[string][]string{}
+				testSuite := sar.ActionProto.GetTestAction().GetTestSuite()
+				for _, t := range testSuite.GetTests() {
+					appendProperties(properties, t, nil)
```
michelle192837, line 532: Hmm, it looks like we collect target result properties as well, but not for this (just for cell.Metrics instead; see targetResultNumerics() internally). I think it's worth a TODO to check if we need to include the properties from the target in addition to its test cases.

michelle192837 (commented):
Mostly good, thanks! Left some comments for cleanup and consideration.

pkg/updater/resultstore/resultstore.go
michelle192837: Think this is still true; checkOrder == annotations, so instead of creating and appending to a new slice we can remove checkOrder altogether and just use annotations in this loop.

michelle192837 (commented):
Last nit: Please squash commits once the last comment is addressed, then it should be good to go. Thanks!

pkg/updater/resultstore/resultstore.go
bryan-lou: Resolved

michelle192837 (approved):

PR #1261, opened 2023-10-03
Filter test methods according to properties and match/unmatch regex

bryan-lou:


pkg/updater/resultstore/resultstore.go
```diff
@@ -408,6 +410,23 @@ func processGroup(tg *configpb.TestGroup, group *invocationGroup) *updater.Infla
 			}
 		}
 
+		// cells := map[string][]updater.Cell{}
```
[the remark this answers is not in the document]
bryan-lou: I think we already have `groupedCells`, so we don't need this

pkg/updater/resultstore/resultstore.go
```diff
@@ -499,6 +626,19 @@ func testMethodLimit(tg *configpb.TestGroup) int {
 	return testMethodLimit
 }
 
+func testMethodRegex(tg *configpb.TestGroup) (matchMethods *regexp.Regexp, unmatchMethods *regexp.Regexp, matchMethodsErr error, unmatchMethodsErr error) {
+	if tg == nil {
+		return nil, nil, nil, nil
+	}
+	if m := tg.TestMethodMatchRegex; m != "" {
```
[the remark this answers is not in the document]
bryan-lou: Done

pkg/updater/resultstore/resultstore.go
```diff
@@ -521,7 +661,7 @@ func mapStatusToCellResult(testCase *resultstorepb.TestCase) statuspb.TestStatus
 
 // processTestResults iterates through a list of test results and adds them to
 // a map of groupedcells based on the method name produced
-func processTestResults(tg *config.TestGroup, testResults []*resultstorepb.Test, cell updater.Cell, targetID string, sar *singleActionResult, testMethodLimit int, groupedCells map[string][]updater.Cell) {
+func processTestResults(tg *config.TestGroup, groupedCells map[string][]updater.Cell, testResults []*resultstorepb.Test, sar *singleActionResult, cell updater.Cell, targetID string, testMethodLimit int, filtered bool) {
```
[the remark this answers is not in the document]
bryan-lou: Removed

pkg/updater/resultstore/resultstore.go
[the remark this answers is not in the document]
michelle192837: Ah, good call on that, thanks!

To clarify my original comment, I'm talking about the section right below it as well. This section adds an error (that we generate) to the grid to let the user know that their regexes didn't compile. This is a late time to surface the warning though; instead we should surface it in the configuration validation test.

pkg/updater/resultstore/resultstore.go
```diff
@@ -521,7 +657,7 @@ func mapStatusToCellResult(testCase *resultstorepb.TestCase) statuspb.TestStatus
 
 // processTestResults iterates through a list of test results and adds them to
 // a map of groupedcells based on the method name produced
-func processTestResults(tg *config.TestGroup, testResults []*resultstorepb.Test, cell updater.Cell, targetID string, sar *singleActionResult, testMethodLimit int, groupedCells map[string][]updater.Cell) {
+func processTestResults(tg *config.TestGroup, groupedCells map[string][]updater.Cell, testResults []*resultstorepb.Test, sar *singleActionResult, cell updater.Cell, targetID string, testMethodLimit int) {
```
michelle192837, line 660: nit: Any reason this is switching the args around?

pkg/updater/resultstore/resultstore.go
```diff
@@ -499,6 +625,16 @@ func testMethodLimit(tg *configpb.TestGroup) int {
 	return testMethodLimit
 }
 
+func testMethodRegex(tg *configpb.TestGroup) (matchMethods *regexp.Regexp, unmatchMethods *regexp.Regexp, matchMethodsErr error, unmatchMethodsErr error) {
```
michelle192837, line 628: Think it's worth adding some simple test cases for this.

pkg/updater/resultstore/resultstore.go
```diff
@@ -458,6 +480,110 @@ func processGroup(tg *configpb.TestGroup, group *invocationGroup) *updater.Infla
...
+			if diff {
+				continue
+			}
+		}
+		properties[key] = values[0]
+	}
+}
+
+// appendProperties from result and its children into a map, optionally filtering to specific matching keys.
+func appendProperties(properties map[string][]string, result *resultstorepb.Test, match map[string]bool) {
```
michelle192837, line 543: Think it's worth adding some simple test cases for this.

pkg/updater/resultstore/resultstore.go
```diff
@@ -458,6 +480,110 @@ func processGroup(tg *configpb.TestGroup, group *invocationGroup) *updater.Infla
...
+		if miss {
+			continue
+		}
+		out = append(out, r)
+	}
+	return out
+}
+
+// fillProperties reduces the appendProperties result to the single value for each key or "*" if a key has multiple values.
+func fillProperties(properties map[string]string, result *resultstorepb.Test, match map[string]bool) {
```
michelle192837, line 516: Think it's worth adding some simple test cases for this.

pkg/updater/resultstore/resultstore.go
bryan-lou: I think we can skip this since `fillProperties` calls this and already include tests there to cover the code and logic here

pkg/updater/resultstore/resultstore.go
bryan-lou: Done

pkg/updater/resultstore/resultstore.go
bryan-lou: Just personal preference; I prefer having the lists, map and struct in front of primitive args 😸

pkg/updater/resultstore/resultstore_test.go
```diff
@@ -1477,6 +2411,55 @@ func TestGetTestResults(t *testing.T) {
 	}
 }
 
+func TestMethodRegex(t *testing.T) {
```
michelle192837, line 2414: nit: Think this should also cover a case with invalid regexes.

michelle192837 (commented):
Mostly looks good, thanks!

Left a comment, and also if you could squash commits that would be great. Thanks!

michelle192837 (approved):
Awesome, thanks! ^^

PR #1264, opened 2023-10-30
Add links for alerts, and utilities to process link templates.

michelle192837:
The frontend currently processes links, but any components that want to use links (like Summarizer or downstream components) can't really do so. Replicate the logic for link-template parsing so we can make use of the user-specified links instead of hard-coded ones.

### DIFF
--- a/pkg/summarizer/BUILD.bazel
+++ b/pkg/summarizer/BUILD.bazel
@@ -22,6 +22,7 @@ go_library(
         "//pkg/summarizer/analyzers:go_default_library",
         "//pkg/summarizer/common:go_default_library",
         "//pkg/tabulator:go_default_library",
+        "//util:go_default_library",
         "//util/gcs:go_default_library",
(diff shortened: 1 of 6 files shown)


sultan-duisenbay (approved):
qq - why are we checking and logging the new links instead of overwriting the existing ones in the failing test summaries? This part is unclear to me.

Otherwise, looks good to me functionality wise.

/hold

michelle192837:
> qq - why are we checking and logging the new links instead of overwriting the existing ones in the failing test summaries? This part is unclear to me.
> 
> Otherwise, looks good to me functionality wise.
> 
> /hold

Good question, thanks! I don't want to break existing use of the links until I verify that everything's working as expected, especially since we haven't touched this in a while (and I suspect I'm missing some invisible assumption about how this works for some tabs). I expect to follow up with one or two more changes before enabling (probably including a flag to flip this behavior for the Summarizer).

PR #1265, opened 2023-11-07
Fix lint issues from code import

bryan-lou:


michelle192837:
Thanks! 

/lint

pkg/updater/resultstore/resultstore.go
michelle192837, line 1: nit: target.status_attributes is deprecated in the target proto, it looks like? (https://github.com/googleapis/googleapis/blob/master/google/devtools/resultstore/v2/target.proto#L55C8-L55C8) Think we should replace that with the recommended field instead.

michelle192837 (commented):
Mostly looks good, left some comments (and the 'lint' command picked up another issue as well). Should be fine once those are addressed, thanks!

michelle192837:
Once the unit tests are fixed and the commits are squashed I think it's good to go!

PR #1253, opened 2023-09-13
feat: #resultstore - add cell metrics

sultan-duisenbay:
Add a feature to  calculate the cell metrics from properties as well as the duration in minutes.

Includes unit tests

PR #1267, opened 2023-11-16
Add cell metrics to resultstore

bryan-lou:
Finishing up Sultan's PR: #1253

### DIFF
--- a/pkg/updater/gcs.go
+++ b/pkg/updater/gcs.go
@@ -526,8 +526,11 @@ func overallCell(result gcsResult) Cell {
 	return c
 }
 
-// ElapsedKey is the key for the test duration metric.
+// ElapsedKey is the key for the target duration metric.
 const ElapsedKey = "test-duration-minutes"
+
+// TestMethodsElapsedKey is the key for the test results duration metric.
+const TestMethodsElapsedKey = "test-methods-duration-minutes"
 
 // setElapsed inserts the seconds-elapsed metric.
 func setElapsed(metrics map[string]float64, seconds float64) map[string]float64 {
--- a/pkg/updater/resultstore/resultstore.go
+++ b/pkg/updater/resultstore/resultstore.go
@@ -508,6 +508,8 @@ func processGroup(tg *configpb.TestGroup, group *invocationGroup) *updater.Infla
 					}
 				}
 
+				cell.Metrics = calculateMetrics(sar)
+
 				// TODO (@bryanlou) check if we need to include properties from the target in addition to test cases
 				properties := map[string][]string{}
 				testSuite := sar.ActionProto.GetTestAction().GetTestSuite()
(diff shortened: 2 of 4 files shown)


pkg/updater/resultstore/resultstore.go
```diff
@@ -536,6 +538,59 @@ func processGroup(tg *configpb.TestGroup, group *invocationGroup) *updater.Infla
...
+			}
+		} else {
+			testResultProperties(properties, test.GetTestSuite())
+		}
+	}
+}
+
+// testResultDuration calculates the overall duration of a target, converts
+// milliseconds to minutes and returns the duration.
+func testResultDuration(target *resultstorepb.Target) int64 {
```
michelle192837, line 581: Comment from https://github.com/GoogleCloudPlatform/testgrid/pull/1253 still applies; I think this should return a time.Duration, or should specify milliseconds (not just in the comment, but in the function name or similar).

It also looks like this no longer returns the duration for the target's test results, just the target?

michelle192837 (commented):
Overall: Please note the PR this is based on when carrying over someone else's change, and credit the original author.

In this case, I might include 2 commits: the one you're basing it off of, and a commit with your changes on top of that, so it's easier to review.

pkg/updater/resultstore/resultstore.go
bryan-lou: Fixed!

pkg/updater/gcs.go
```diff
@@ -529,6 +529,9 @@ func overallCell(result gcsResult) Cell {
 // ElapsedKey is the key for the test duration metric.
 const ElapsedKey = "test-duration-minutes"
 
+// TestMethodsElapsedKey is an automatically-added metric for sponge.
```
michelle192837, line 532: nit: Would match this to "ElapsedKey" above for consistency.

pkg/updater/resultstore/resultstore.go
```diff
@@ -538,18 +538,22 @@ func processGroup(tg *configpb.TestGroup, group *invocationGroup) *updater.Infla
...
 	testResultProperties(properties, sar.ActionProto.GetTestAction().GetTestSuite())
 	numerics := updater.Means(properties)
+	targetElapsed := sar.TargetProto.GetTiming().GetDuration().AsDuration()
+	if targetElapsed > 0 {
+		numerics[updater.ElapsedKey] = float64(targetElapsed.Milliseconds()) / 60000
+	}
 
 	if dur := testResultDuration(sar.ActionProto.GetTestAction().GetTestSuite()); dur > 0 {
-		numerics[updater.ElapsedKey] = float64(dur) / 60000
+		numerics[updater.TestMethodsElapsedKey] = float64(dur.Milliseconds()) / 60000
```
michelle192837, line 553: nit: I think you don't need to do milliseconds conversion, you can just take Minutes() instead?

pkg/updater/resultstore/resultstore_test.go
michelle192837, line 1: I think the unit tests need an update; it looks like some of them are meant to test picking up durations, but don't actually test that now.

pkg/updater/gcs.go
bryan-lou: done

pkg/updater/resultstore/resultstore_test.go
bryan-lou: updated!

pkg/updater/gcs.go
```diff
@@ -526,9 +526,12 @@ func overallCell(result gcsResult) Cell {
 	return c
 }
 
-// ElapsedKey is the key for the test duration metric.
+// ElapsedKey is the key for the test results duration metric.
```
michelle192837, line 529: nit: test-duration-minutes should be the test / target duration.

The one below, for consistency, should be "test-methods-duration-minutes", for the test methods / test results.

michelle192837: Sorry to nitpick, I just think consistency with the codebase will be helpful in the long run.

Can this get updated to:

```
// ElapsedKey is the key for the target duration metric.
const ElapsedKey = "test-duration-minutes"

// TestMethodsElapsedKey is the key for the test results duration metric.
const TestMethodsElapsedKey = "test-methods-duration-minutes"
```

Thanks! Looks good to go after that.

bryan-lou: Fixed! @michelle192837

michelle192837 (approved):

PR #1268, opened 2023-11-28
Add retry when snapshotting configuration.

michelle192837:
Adding a quick retry to mitigate issues where the configuration is (temporarily) not available on startup.

Added a unit test for this as well (probably could be cleaner, but it should verify that the retry actually works).

ETA: The actual change to add a retry is in config_snapshot.go. Everything else is an update to add a unit test for the retry, and enhance the storage fake to allow that (with a lock to prevent data races on map reads/writes for the fake Opener).

### DIFF
--- a/config/snapshot/config_snapshot_test.go
+++ b/config/snapshot/config_snapshot_test.go
@@ -73,7 +73,7 @@ func TestObserve_OnInit(t *testing.T) {
 			defer cancel()
 
 			client := fakeClient()
-			client.Opener[*path] = fake.Object{
+			client.Opener.Paths[*path] = fake.Object{
 				Data: string(mustMarshalConfig(test.config)),
 				Attrs: &storage.ReaderObjectAttrs{
 					Generation: test.configGeneration,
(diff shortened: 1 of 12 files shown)


michelle192837:
Ah, there's a legitimate race in the new fake logic I believe. Taking a look at that.

michelle192837:
Test should be fixed, ready for review!

config/snapshot/config_snapshot_test.go
```diff
@@ -108,6 +108,96 @@ func TestObserve_OnInit(t *testing.T) {
 	}
 }
 
+func TestObserve_OnInitRetry(t *testing.T) {
+	tests := []struct {
+		name             string
+		config           *configpb.Configuration
+		configGeneration int64
```
chases2, line 115: I don't think "configGeneration" is actually a parameter of this test. Consider just hard-coding it when constructing your fake.

config/snapshot/config_snapshot_test.go
```diff
@@ -108,6 +108,96 @@ func TestObserve_OnInit(t *testing.T) {
...
+			name: "Returns error if config isn't present on retry",
+			config: &configpb.Configuration{
+				Dashboards: []*configpb.Dashboard{
+					{
+						Name: "dashboard",
+					},
+				},
+			},
+			configGeneration: 1,
+			expectInitial: &configpb.Dashboard{
```
chases2, line 147: Do we expect an initial config in this test case? It looks like we'd expect just an error. "openOnRetry" is false, so it's just going to keep failing, like if the file just isn't there.

This _happens_ to succeed because your test code returns early after "expectError" and never actually checks this. It's fine to return early so we don't wait 30 seconds for a config we don't expect, but please remove it from the test setup array here.

config/snapshot/config_snapshot_test.go
```diff
@@ -108,6 +108,96 @@ func TestObserve_OnInit(t *testing.T) {
...
+			}
+			client.Stater[*path] = fake.Stat{
+				Attrs: storage.ObjectAttrs{
+					Generation: test.configGeneration,
+				},
+			}
+
+			snaps, err := Observe(ctx, nil, client, *path, nil)
+
+			if err != nil {
```
chases2, line 182: nit: You want to fail if these don't match, right? So you'll want to either check both cases (including "we expected an error, but didn't get one), or a more complex check like `if (err != nil) != test.expectError`

chases2 (commented):
The code itself looks good (after digging around a bit in this new "retry" library) and the tests demonstrates the API well (ie: you will get a config OR you will get an error, etc), but I think it could be cleaner. Please let me know what you think!

config/snapshot/config_snapshot_test.go
michelle192837: Ah yup, fixed it up to check both cases. (iirc we do have unit tests in other components that do this if/else style when expecting errors.)

config/snapshot/config_snapshot_test.go
michelle192837: Done, thanks!

michelle192837:
> The code itself looks good (after digging around a bit in this new "retry" library) and the tests demonstrates the API well (ie: you will get a config OR you will get an error, etc), but I think it could be cleaner. Please let me know what you think!

Thanks for catching these! Agree on the suggestions, implemented them and updated the logic in the fake to be clearer (pretty sure I hit some issues on the boolean conditions).

chases2 (approved):
Looks great! Thanks for taking the extra time on the tests. :D

Feel free to merge when ready; I only see one tiny nit after the refactor.

/hold

config/snapshot/config_snapshot_test.go
```diff
@@ -108,6 +108,87 @@ func TestObserve_OnInit(t *testing.T) {
...
+		{
+			name: "Reads config on retry",
+			config: &configpb.Configuration{
+				Dashboards: []*configpb.Dashboard{
+					{
+						Name: "dashboard",
+					},
+				},
+			},
+			expectInitial: &configpb.Dashboard{
```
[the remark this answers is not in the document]
michelle192837: Done, thanks! o7

PR #1270, opened 2024-02-14
Fix ResultStore updater to only update relevant groups.

michelle192837:
Looks like this accidentally reused the logic for the GCS updater, so fixing that. Added unit tests as well.

### DIFF
--- a/pkg/updater/resultstore/resultstore_test.go
+++ b/pkg/updater/resultstore/resultstore_test.go
@@ -3540,3 +3540,68 @@ func TestCalculateMetrics(t *testing.T) {
 		})
 	}
 }
+
+func TestShouldUpdate(t *testing.T) {
+	cases := []struct {
+		name         string
+		tg           *configpb.TestGroup
+		clientExists bool
+		want         bool
+	}{
+		{
+			name: "nil",
+			want: false,
+		},
+		{
+			name: "test group config only",
+			tg: &configpb.TestGroup{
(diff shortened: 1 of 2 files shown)


pkg/updater/resultstore/resultstore_test.go
```diff
@@ -3540,3 +3540,68 @@ func TestCalculateMetrics(t *testing.T) {
...
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			// Create a fake client if specified.
+			var dlClient *DownloadClient
+			if tc.clientExists {
+				dlClient = &DownloadClient{client: &fakeClient{}}
+			}
+			if got := shouldUpdate(logrus.WithField("name", tc.name), tc.tg, dlClient); tc.want != got {
+				t.Errorf("shouldUpdate(%v, clientExists = %t) differed; got %t, want %t", tc.tg, tc.clientExists, got, tc.want)
```
airbornepony, line 3603: I usually reserve "differ" for cases where I have an actual diff to report. Your preference though.
```suggestion
				t.Errorf("shouldUpdate(%v, clientExists = %t) got %t, want %t", tc.tg, tc.clientExists, got, tc.want)
```

airbornepony (approved):

pkg/updater/resultstore/resultstore_test.go
michelle192837: That's fair, thanks! Updated ^^

PR #1271, opened 2024-02-15
Add custom column headers to summarizer

prernakakkar-google:


pkg/summarizer/summary.go
```diff
@@ -858,15 +858,16 @@ func failingTestSummaries(rows []*statepb.Row, template *configpb.LinkTemplate,
...
-			HotlistIds:         alert.HotlistIds,
-			EmailAddresses:     alert.EmailAddresses,
+			BuildLink:           alert.BuildLink,
+			BuildLinkText:       alert.BuildLinkText,
+			BuildUrlText:        alert.BuildUrlText,
+			LinkedBugs:          row.Issues,
+			FailTestLink:        buildFailLink(alert.FailTestId, row.Id),
+			LatestFailTestLink:  buildFailLink(alert.LatestFailTestId, row.Id),
+			Properties:          alert.Properties,
+			CustomColumnHeaders: alert.CustomColumnHeaders,
```
michelle192837, line 868: Note that 'alert' has to pull info from somewhere first; this is set in the Updater in alertRow() (https://github.com/GoogleCloudPlatform/testgrid/blob/main/pkg/updater/updater.go#L1281).

util/links.go
```diff
@@ -36,7 +36,7 @@ var (
 
 	// Substitute a particular custom column value for this token.
 	// TODO: Handle start/end columns (for custom columns and build IDs).
-	CustomColumnRe = regexp.MustCompile(`<custom-\d+>`)
+	CustomColumnRe = regexp.MustCompile(`<custom-\w+>`)
```
michelle192837, line 39: For consistency, I think we still want to support <custom-N> (e.g. `<custom-0>` for the value of the first custom column header), so I don't think this regex should change.

That means that for getting the column headers, you'll need to compare to the column headers in configuration. This should be available in Summarizer, for instance, through updateTab()'s `group` argument (which contains the TestGroup configuration with the custom column headers), which can be used in failingTestSummaries() to pull the custom column header info.

pb/state/state.proto
```diff
@@ -88,6 +88,9 @@ message AlertInfo {
   // Dynamic email list, route email alerts to these instead of the configured
   // defaults.
   repeated string email_addresses = 15;
+
+  // Maps (custom column headers name):(custom column headers value) for arbitrary alert properties.
+  map<string, string> custom_column_headers = 16;
```
michelle192837, line 93: (Note: this applies to summary.proto too).

Since this is an alert, we're interested in column headers from two columns; the first column to fail, and the last column to pass. I think you either want to clarify which column this is pulled from (probably the first column to fail), or add fields for both (similar to `fail_build_id` and `pass_build_id` above). I'd recommend the former.

michelle192837 (commented):
Looks like a good start, thanks! Left some comments, please let me know if you have any questions ^^

pkg/updater/updater_test.go
```diff
@@ -4753,10 +4790,11 @@ func TestAlertRow(t *testing.T) {
 				CellIds:  []string{"no", "no again", "very wrong", "yes", "hi", "hello"},
 			},
 			failOpen: 3,
-			expected: alertInfo(3, "no", "very wrong", "no", nil, columns[2], columns[0], columns[3], false),
+			expected: alertInfo(3, "no", "very wrong", "no", nil, columns[2], columns[0], columns[3], false, customColumnHeaders),
```
michelle192837, line 4793: nit: I'd test the custom column headers on 1 or 2 cases, instead of adding it for all cases, just to keep the other test cases a bit clearer.

michelle192837 (commented):
Thanks! This looks good, I just have two small things to address:

- The "nit:" comments on the review
- If you can squash your commits together into 1 commit, that'd be great. Thanks!

Once that's done and I approve, the PR'll submit, and I can make a new release for import. ^^

michelle192837 (approved):
Thanks!

PR #1272, opened 2024-02-28
Clarify message for missing podinfo.json.

michelle192837:
TestGrid expects that podinfo.json is uploaded by the time that
finished.json is, and will either report it as a running/pending
status if finished.json isn't uploaded yet, or as a failure if it is.

The message that appears on the <job>.Pod cell isn't clear about this
case, so updating the message to include it. (The previous case
mentioned, where Prow's GCS reporter isn't enabled, still applies, but
podinfo.json can fail to be uploaded for other reasons as well).

### DIFF
--- a/util/gcs/read.go
+++ b/util/gcs/read.go
@@ -44,9 +44,9 @@ type PodInfo struct {
 
 const (
 	// MissingPodInfo appears when builds complete without a podinfo.json report.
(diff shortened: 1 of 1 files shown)


airbornepony (approved):

chases2 (approved):

michelle192837:
/hold cancel

Updated description/commit as well. Thanks folks!

PR #1273, opened 2024-02-28
Fixing a panic when recording column headers in alerts.

michelle192837:
A recent update to Summarizer records column header values (specifically, the 'extra' or 'custom' column header values) when recording metadata about alerted rows. We missed a case where, if there are fewer values recorded than the current number of custom column headers configured for the tab, we hit an out of index issue. (This is a legitimate case that can happen when new custom column headers are added).

Quick-fixed this by checking the length of column header values before recording them, and an additional unit test to verify this case. (We made this change internally, and I've verified it works already).

### DIFF
--- a/pkg/updater/updater.go
+++ b/pkg/updater/updater.go
@@ -1342,6 +1342,13 @@ func alertRow(cols []*statepb.Column, row *statepb.Row, failuresToOpen, passesTo
 		compressedIdx++
 
 		for i := 0; i < len(columnHeader); i++ {
+			if i >= len(col.Extra) {
+				logrus.WithFields(logrus.Fields{
+					"started":                 time.Unix(0, int64(col.GetStarted()*float64(time.Millisecond))),
+					"additionalColumnHeaders": col.GetExtra(),
+				}).Trace("Insufficient column header values to record.")
+				break
+			}
 			if columnHeader[i].Label != "" {
 				customColumnHeaders[columnHeader[i].Label] = col.Extra[i]
 			} else if columnHeader[i].Property != "" {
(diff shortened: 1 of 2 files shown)


pkg/updater/updater.go
```diff
@@ -1342,6 +1342,10 @@ func alertRow(cols []*statepb.Column, row *statepb.Row, failuresToOpen, passesTo
 		compressedIdx++
 
 		for i := 0; i < len(columnHeader); i++ {
+			if i >= len(col.Extra) {
+				logrus.WithField("started", time.Unix(0, int64(col.GetStarted()*float64(time.Millisecond)))).WithField("additionalColumnHeaders", col.GetExtra()).Trace("Insufficient column header values to record.")
```
listx, line 1346: IIRC logrus lets you do multiple fields at once...

Found example: https://github.com/kubernetes/test-infra/blob/9cd5611d0a632c3f04d2c03cc56d58ebd887b9b7/prow/github/webhooks.go#L74

airbornepony (approved):
/hold for others

pkg/updater/updater.go
michelle192837: Ah yup, I just usually do it at 3+ fields.

This is pretty long though, so updated!

listx:
@michelle192837 Also, are you sure there is no additional context for this change for your future self/others? I find it very rare that a commit doesn't need a corresponding commit description (the usual suspects are typically just typofixes).

michelle192837:
> @michelle192837 Also, are you sure there is no additional context for this change for your future self/others? I find it very rare that a commit doesn't need a corresponding commit description (the usual suspects are typically just typofixes).

Good point, sorry about that. Updated the description with more details, thanks!

listx:
/lgtm

Thanks

michelle192837:
/hold cancel

Thanks!

PR #1274, opened 2024-03-01
Add ability to query a single job from Prow in ResultStore.

michelle192837:
Our ResultStore implementation at the moment is limited to Prow results from ResultStore (e.g. has "label:prow"). This is fine for our current assumptions, but if results for multiple Prow jobs are uploaded to the same project (a common use case), then all the results get processed in TestGrid on the same tab, which we want to avoid for users and development.

Instead, allow a very simple query (only `target:"<job name>"`), so it's possible to specify a particular job for a tab instead of the results for everything uploaded from Prow.


pkg/updater/resultstore/query.go
```diff
@@ -0,0 +1,63 @@
...
+	switch {
+	case key == "target":
+		return fmt.Sprintf(`id.target_id="%s"`, val), nil
+	default:
+		return "", fmt.Errorf("unrecognized atom key %q", key)
+	}
+}
+
+var (
+	queryRe = regexp.MustCompile(`target:"?.*"?`)
```
airbornepony, line 47: Seems this regexp permits patterns like
`target:"foo`
`target:foo"`
`target:"foo", "bar"`
etc.

Do you need it to be more defensive? What is the source of these strings?

pkg/updater/resultstore/query.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+func translateQuery(simpleQuery string) (string, error) {
+	if simpleQuery == "" {
+		return "", nil
+	}
+	// For now, we expect a query with a single atom, with the exact form `target:"<target>"`
+	if !queryRe.MatchString(simpleQuery) {
```
airbornepony, line 55: MatchString matches a substring--"target:" could appear anywhere. Should anchor if that is the intent.

airbornepony (changes requested):

pkg/updater/resultstore/query.go
michelle192837: Thanks for the reminder! Updated, and added a test case to catch it.

pkg/updater/resultstore/query.go
michelle192837: It doesn't have to be, but for now it probably should be. The source of these is either user-defined (so someone can misconfigure it, we just want to discourage it), or automatically-generated TestGrid groups (like groups created from Prow annotations or the like; in this case, for auto-generation when we know a Prow job is using a ResultStore source).

Tightened up the regular expression, thanks for catching this!

PR #1276, opened 2024-03-05
Capture configured target stats from ResultStore.

michelle192837:
Forgot to add field masks for fetching configured target stats from ResultStore. Without this, status and timing info don't get fetched when downloading results.


chases2 (approved):

michelle192837:
Thanks Sean! ^^

PR #1278, opened 2024-03-14
Use SearchConfiguredTargets when searching targets.

michelle192837:
If searching for a target (as opposed to labels or other identifiers), we want to search for those results using a difference method (SearchConfiguredTargets, rather than SearchInvocations). Add the search method and use it if the query contains a target atom (and update the Prow label accordingly for this search).

Moved the query code into a separate package so it can be used in config validation without a circular dependency.

### DIFF
--- a/config/BUILD.bazel
+++ b/config/BUILD.bazel
@@ -13,6 +13,7 @@ go_library(
     visibility = ["//visibility:public"],
     deps = [
         "//pb/config:go_default_library",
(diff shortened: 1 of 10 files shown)


michelle192837:
/hold

Making one more small update.

michelle192837:
This is pushed to the wrong branch, closing and re-opening with a new one.

PR #1289, opened 2024-09-16
Specify auto-project detection for PubSub.

michelle192837:
In pubsub v1.32, a breaking change made it so specifying a blank project results in an error, rather than automatically detecting the project as it did before. Instead, specify the sentinel value for detecting the project automatically.

See
https://github.com/googleapis/google-cloud-go/releases/tag/pubsub%2Fv1.32.0.

Note that an update a while back upgraded the pubsub module version from v1.22.2 to v1.33.0.

Verified locally that this change fixes some startup failures that exist on the current images. (This should be fixed the next time images are built and the autobump includes them in configs).

### DIFF
--- a/cmd/summarizer/main.go
+++ b/cmd/summarizer/main.go
@@ -96,7 +96,7 @@ func gcsFixer(ctx context.Context, projectSub string, configPath gcs.Path, tabPr
 		return nil, errors.New("malformed project/subscription")
 	}
 	projID, subID := parts[0], parts[1]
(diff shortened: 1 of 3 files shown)


michelle192837:
/hold

One moment, updating tabulator and summarizer as well

michelle192837:
/hold cancel

Updated all three files and verified that they work with the change.

chases2 (approved):
LGTM. Thanks!

PR #1290, opened 2024-09-17
Subscribe to gs://kubernetes-ci-logs updates.

michelle192837:
Includes two changes:
- Add a canary and prod subscription on gs://kubernetes-ci-logs to the PubSub setup script. (And run it so the new subscriptions are registered on k8s-testgrid. This should eventually get replaced with Terraform setup, but that's probably better to do later if/when we migrate to a separate setup for Kubernetes).
- Add the subscriptions to the Updater config, so the Updater responds to updates on gs://kubernetes-ci-logs (same as it does for gs://kubernetes-jenkins).

Ref https://github.com/kubernetes/test-infra/issues/33386
(That will be fixed once these changes make it to prod).


michelle192837:
Update: Now that some groups in k8s-infra-prow pull from gs://kubernetes-ci-logs, I can confirm that this subscription works in canary. (This change will update the canary config to match, and will register the subscription in prod too).

nathanperkins (approved):

PR #1296, opened 2024-11-08
Allow label/target search for ResultStore sources.

michelle192837:
Instead of requiring `label:"prow"` in queries, allow any number of label or target atoms in a query. (This still limits the queries we accept, but it should work for most simple uses of ResultStore sources).

Test: Unit tests, and verified with a local instance of TestGrid that a ResultStore source config will pick up results from ResultStore correctly.

### DIFF
--- a/pkg/updater/resultstore/query/query.go
+++ b/pkg/updater/resultstore/query/query.go
@@ -17,47 +17,76 @@ limitations under the License.
 package query
 
 import (
+	"errors"
 	"fmt"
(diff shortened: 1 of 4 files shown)


michelle192837:
Found an issue with a growing number of goroutines that didn't pop up on a simple search but does with a more complex search. I need to fix that before this is good to roll out.

michelle192837:
Obsolete

PR #1313, opened 2025-08-13
ResultStore: Update Invocation and Invocation ID

madziejm:
Please review the two commits attached.


resultstore/client.go
```diff
@@ -213,6 +213,19 @@ func (t Targets) Create(id string, target Target) (string, error) {
 	return tgt.Name, nil
 }
 
+// Update a pre-existing target.
+func (t Targets) Update(name, id string, target Target, fields ...string) error {
+	pbTarget := target.To()
```
michelle192837, line 218: nit: I think we usually use `<variable>Proto` for protobufs.

